PR #2, opened 2022-02-11
Make crate compatible with stable Rust

phil-opp:
Hi, I saw your great post about the shortcomings of `select!` and your proposed `Stream::merge` operator. I would like to try to replace `select!`s with the `merge` operator in some projects that are targeting stable Rust. To make this possible, this PR tries to remove the nightly requirement of this crate.

Out of the three uses of nightly features, two can be replaced by inlining the respective unstable functions (their implementations use no nightly features). The third feature, `array_methods`, is only used in an internal function that appears to be unused. Given that this crate is for experimentation, I still kept it, but moved it behind a new `unstable` cargo feature. I also updated the crate docs with a note that some features are only available on nightly. While this statement isn't true right now, it allows you to freely experiment with unstable features in future versions of this crate.

Let me know what you think of this approach. And thanks for your work to make async Rust better!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ authors = [
 ]
 
 [features]
+unstable = []
 
 [dependencies]
 futures-core = "0.3.16"
--- a/src/join/array.rs
+++ b/src/join/array.rs
@@ -68,7 +68,11 @@ where
             use core::mem::MaybeUninit;
 
             // Create the result array based on the indices
-            let mut out: [MaybeUninit<F::Output>; N] = MaybeUninit::uninit_array();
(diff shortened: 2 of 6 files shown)


yoshuawuyts (approved):
This is great, thanks so much!

yoshuawuyts:
I've released `v0.2.1` which includes this fix. Please be aware though that https://github.com/yoshuawuyts/futures-concurrency/issues/3 is a thing, and calling `merge` on tuples is currently broken at least for tuples of length 3. Right now it's a manual impl on tuples, but it probably should be a macro expansion. Help moving to a macro would be greatly appreciated.

PR #4, opened 2022-02-28
Fix merge implementation on tuples

### LINKED ISSUE #3: stream tuple merge is broken
Not sure what the issue is, but https://github.com/yoshuawuyts/futures-concurrency/pull/2 flagged broken CI. Tested it locally, and it only happens with tuples - not with arrays.

phil-opp:
Add a recursive `poll_in_order!` macro and use it for implementing `Merge` on tuples. This fixes merging for 3-tuples, which was broken because the previous `poll_next_in_order` function did no longer poll `s2` after `s0` finished: https://github.com/yoshuawuyts/futures-concurrency/blob/53266472d28074054b064372440714bc630cd21b/src/merge/tuple.rs#L156

To utilize the new macro, this PR also adds a `Stream` implementation for 4-tuples. It still requires a lot of boilerplate, but at least the repetitive matching on `poll_next` is now autogenerated.

I also tried to create a macro to autogenerate the `MergeX` structs and trait impl blocks, but I didn't find a way to permutate the list of streams in a macro.

Fixes #3

### DIFF
--- a/src/merge/tuple.rs
+++ b/src/merge/tuple.rs
@@ -6,6 +6,22 @@ use futures_core::Stream;
 use std::pin::Pin;
 use std::task::{Context, Poll};
 
+macro_rules! poll_in_order {
+    ($cx:expr, $stream:expr) => { $stream.poll_next($cx) };
+    ($cx:expr, $stream:expr, $($next:tt),*) => {{
+        let mut pending = false;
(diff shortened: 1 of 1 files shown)


yoshuawuyts (approved):
This is great, thanks so much!

yoshuawuyts:
published as `v2.0.2`

PR #8, opened 2022-07-05
Wake futures once instead of every time

yoshuawuyts:
WIP with @eholk


yoshuawuyts:
Closing in favor of #50

eholk:
We'll still need to do this for arrays too, of course, but it's probably easier to reuse the existing working implementation on vecs.

Issue #18, opened 2022-11-03
Implement `stream::Merge` for tuple lengths up to 12


yoshuawuyts:
From Oli, we might be able to do something like this:

```rust
let r = random(24);
for i in 0..=4 {
    if i == r/6 {
        poll(s0)
    } else if i == (r + 1)/6 {
        poll(s1)
    } ...
}
```

Issue #16, opened 2022-11-01
`Join` futures should exist on a path


yoshuawuyts:
- rename them to `Join0`, `Join1`, etc.
- expose them as `pub(crate)` on an actual path
- this is to make debuggers happy according to mw

michaelwoerister:
Yep, debuggers often need to refer to types by their fully qualified name and the way things are set up right now (i.e., the definition being inside the initializer of an unnamed `const`), the fully qualified name might contains things that are implementation details of `rustc` and thus have no stability guarantee.

PR #19, opened 2022-11-04
implement `stream::Merge` for tuples up to length 12

yoshuawuyts:
Closes https://github.com/yoshuawuyts/futures-concurrency/issues/18

The key insight for this patch was provided by @oli-obk. Rather than iterating through all possible stream permutations, this approach makes it so we have a randomized starting point, and then go through all streams in a fixed order.

Over N-iterations, that still has a "fair" ordering between items. But unlike our prior approach, we can express this entirely using conditionals, which enables auto-generating them. I've added tests for `{2,3,4}`-arity merges, and all seem to pass. Providing a good degree of confidence this patch should continue to work on all existing code. But we now support up to `12`-arity stream merges.

As an added bonus, the generated code should be substantially smaller, which may provide some size reductions as well.

## Notes for stdlib

This code can probably be simplified a bit when we propose it for inclusion in the stdlib. [RFC 3086](https://rust-lang.github.io/rfcs/3086-macro-metavar-expr.html) provides more tools for declarative macros, which are not yet available on stable. But we should be able to rely on those within the stdlib.

---

r? @phil-opp - this shouldn't break any code you have, but I'm tagging you to get an extra pair of eyes on this just in case. Thanks!

### DIFF
--- a/src/stream/merge/tuple.rs
+++ b/src/stream/merge/tuple.rs
@@ -22,20 +22,8 @@ const fn permutations(mut num: u32) -> u32 {
 
 /// Calculate the number of tuples currently being operated on.
 macro_rules! tuple_len {
-    // Increment the counter.
-    (@inner $counter:expr, $F:ident, $($rest:ident,)*) => {
-        tuple_len!(@inner $counter + 1, $($rest,)*)
-    };
-
-    // End of recursion.
-    (@inner $counter:expr,) => {
-        $counter
-    };
-
-    // Base condition.
-    ($($F:ident,)*) => {
-        tuple_len!(@inner 0, $($F,)*)
-    }
+    (@count_one $F:ident) => (1);
+    ($($F:ident,)*) => (0 $(+ tuple_len!(@count_one $F))*);
 }
 
 /// Generate the `match` conditions inside the main `poll_next` body. This macro
--- a/src/stream/merge/tuple.rs
+++ b/src/stream/merge/tuple.rs
@@ -7,6 +7,7 @@ use std::pin::Pin;
 use std::task::{Context, Poll};
 
 /// Compute the number of permutations for a number
+/// during compilation.
 const fn permutations(mut num: u32) -> u32 {
     let mut total = 1;
     loop {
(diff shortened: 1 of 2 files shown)


src/stream/merge/tuple.rs
```diff
@@ -6,239 +6,193 @@ use crate::stream::Stream;
...
-impl<T, A, B, C> Stream for Merge3<T, A, B, C>
-where
-    A: Stream<Item = T>,
-    B: Stream<Item = T>,
-    C: Stream<Item = T>,
-{
-    type Item = T;
+        #[derive(Debug)]
+        #[pin_project::pin_project]
+        pub struct $StructName<T, $($F),*>
```
RReverser, line 105: You could extract this as a single generic struct `Merge<Out, Inputs>` outside of macro, and then in the macro you would only need to generate `impl`s like

```rust
impl<T, $($F),*> Stream for $StructName<T, ($($F,)*)> { ... }
```

(note the parentheses around `($($F,)*)` - that is, you would have impls where 2nd generic argument is a tuple)

RReverser: Ah although I'm not 100% sure if `#[pin]` supports tuples.

yoshuawuyts: yeah, the combination of `#[pin]` + https://github.com/yoshuawuyts/futures-concurrency/issues/16 probably make it so having explicitly named unique futures exported as `pub(crate)` probably make for the best choice right now.

If we exported the futures in any public path, I'd probably jump on it. But the plan is to somewhat intentionally keep it all private.

src/stream/merge/tuple.rs
```diff
@@ -6,239 +6,193 @@ use crate::stream::Stream;
...
-
-impl<T, A, B> Merge2<T, A, B>
-where
-    A: Stream<Item = T>,
-    B: Stream<Item = T>,
-{
-    pub(crate) fn new(streams: (A, B)) -> Self {
-        Self { streams }
+/// Calculate the number of tuples currently being operated on.
+macro_rules! tuple_len {
```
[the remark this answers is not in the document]
yoshuawuyts: great shout, thank you!

Admittedly I'm not super comfortable with decl macros, so I appreciate you jumping in to provide a simpler solution. Ty ty!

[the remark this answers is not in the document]
RReverser: I love macros (probably too much heh), feel free to tag me if you have more questions in the future!

yoshuawuyts:
We've had reviews on this, and it'll be a few days before we actually do another release (at least). Going to go forward and merge this to unblock some of the other issues.

PR #26, opened 2022-11-07
expose tuple futures from path

### LINKED ISSUE #16: `Join` futures should exist on a path

yoshuawuyts:
Closes #16, exposes the tuple futures with a unique name from a private path.

r? @michaelwoerister 

## Screenshot

Running `cargo doc --document-private-items` shows this should work as expected:

Capture


yoshuawuyts:
Talked with mw over chat; this looks good, but we'll run a validation pass before creating a new release. Merging this now!

Issue #22, opened 2022-11-04
Move away from `MaybeDone` to sets of outputs


yoshuawuyts:
Right now we're using `MaybeDone` internally to track the state of each future. It can either be: "pending", "containing data", or "done". This is nice from an implementation perspective, but bad for performance. Namely:

1. It creates unnecessary copies, and sometimes also allocations. Once to store the futures in the `MaybeDone` structure. And once to copy the data to the output structure.
2. It creates unnecessarily large memory needs. At any point we may be using up to 2x the amount of memory needed, since we're doing redundant copies.

The solution is to move the container of futures in-line. Create the container of outputs containing `MaybeUninit`. And finally create a separate container to track the state. Once all futures have completed, we can mark the output container as "initialized" and immediately yield that.

poliorcetics:
So @yoshuawuyts talked about this on mastodon, I proposed in response to associate index and relevant generic in the macros like so:

```rust
macro_rules! associated {
    (/* other elements */; $( ($Fi: literal, $Ft: ident) ),+; /* other elements */) => {
        #[test]
        fn test_indexes_are_sorted() {
            let indexes = [$($Fi),+];
            assert!(indexes.is_sorted(), "Indexes are not sorted in call to macro `associated!()`");
        }
        // With the test running in CI you can use the following to get the max index easily:
        const LEN: usize = {
            let indexes = [$($Fi),+];
            let min = match indexes {
                 [min, ..] => min,
            };
            let max = match indexes {
                 [.., max] => max,
            };
            max - min + 1
        } ;
    };
}

**Comment by poliorcetics** [comment]
If you are having names conflicts for testing or such, what you can do is add a `$test_mod_name: ident` member to the macro and put them in 

```rust
#[cfg(test)
mod $test_mod_name {
    use super::*;
    /* put tests here */
}
```

yoshuawuyts:
https://github.com/yoshuawuyts/futures-concurrency/pull/74 is probably the furthest along in setting up the machinery required for the tuples. It just needs one last change to ensure that tuple outputs are written in-place, rather than copied.

poliorcetics:
It can probably be done without using "paste" but I'm being picky 😄 if I find the time to make a PR before the other is merged I'll make a comparison

PR #29, opened 2022-11-07
Remove `MaybeDone` from `impl Join for Vec`

yoshuawuyts:
This is the first step towards #22. Conceptually it moves us from the existing structure, which has all things together in a single enum:

```rust
pub struct Join<Fut: Future> {
    state: Pin<Vec<MaybeDone<Fut>>>>,
}
```

Towards a separated input/output/metadata structure, which has far better locality:

```rust
pub struct Join<Fut: Future> {
    #[pin]
    futures: Vec<Fut>,
    items: Vec<MaybeUninit<<Fut as Future>::Output>>,
    metadata: Vec<Metadata>,
}
```

This requires some unsafe at the end to transmute the `MaybeUninit<T>` to `T`. Which also requires us to mess around with custom drop implementations. But as a result, it should have better locality since we only ever have to write the data once.

---

Because this is the first iteration of a series of PRs, it has a lot more things going on. We introduce a new type: `Metadata`, which tracks the state. Unlike `MaybeDone` which wraps a `T`, `Metadata` is expected to be co-located with the state. Each piece of `Metadata` carries an enum `State`, which tracks what state the futures and output are in.

---

r? @eholk, this ties into some of the earlier work we've been doing. Tagging you so you can take a look if you're interested!


src/future/join/vec.rs
```diff
@@ -1,50 +1,58 @@
...
-        Self { elems }
+    pub(crate) fn new(futures: Vec<Fut>) -> Self {
+        Join {
+            items: std::iter::repeat_with(|| MaybeUninit::uninit())
+                .take(futures.len())
+                .collect(),
+            metadata: std::iter::successors(Some(0), |prev| Some(prev + 1))
+                .take(futures.len())
+                .map(Metadata::new)
+                .collect(),
```
eholk, line 41: It probably doesn't make much difference, but I think this could also be `(0..futures.len()).into_iter().map(Some).map(Metadata::new).collect()`.

src/future/join/vec.rs
```diff
@@ -64,23 +73,82 @@ where
...
+        // Check whether we're all done now or need to keep going.
+        if this.metadata.iter().all(|meta| meta.is_done()) {
+            // Mark all data as "taken" before we actually take it.
+            this.metadata.iter_mut().for_each(|meta| meta.set_taken());
+
+            // SAFETY: we've checked with the metadata that all of our outputs have been
+            // filled, which means we're ready to take the data and assume it's initialized.
+            let items = unsafe {
+                let items = mem::take(this.items);
+                mem::transmute::<_, Vec<Fut::Output>>(items)
```
eholk, line 105: Can you use `assume_init` here?

src/future/join/vec.rs
```diff
@@ -64,23 +73,82 @@ where
...
         }
 
-        if all_done {
-            let mut elems = mem::replace(&mut self.elems, Box::pin([]));
-            let result = iter_pin_mut(elems.as_mut())
-                .map(|e| e.take().unwrap())
-                .collect();
-            Poll::Ready(result)
+        // Check whether we're all done now or need to keep going.
+        if this.metadata.iter().all(|meta| meta.is_done()) {
```
eholk, line 97: You could also probably keep a counter, like `this.pending_futures` and decrement that when one is done. Then you don't have to iterate over all the metadata each time, it's just a check for 0.

But, I'm guessing it's not common to join more than a handful of futures, so maybe the performance doesn't matter too much?

src/future/join/vec.rs
```diff
@@ -64,23 +73,82 @@ where
 {
     type Output = Vec<Fut::Output>;
 
-    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
-        let mut all_done = true;
+    // SAFETY: see https://github.com/rust-lang/rust/issues/104108,
+    // projecting through slices is fine now, but it's not yet guaranteed to
+    // work. We need to guarantee structural pinning works as expected for it to
+    // be provably sound.
```
eholk, line 79: Just a minor style preference, I'd move this inside the poll function because here it reads like a doc comment to me.

src/utils/metadata.rs
```diff
@@ -0,0 +1,105 @@
...
+impl PollState {
+    /// Returns `true` if the poll state is [`Active`].
+    ///
+    /// [`Active`]: PollState::Active
+    #[must_use]
+    fn is_active(&self) -> bool {
+        matches!(self, Self::Active)
+    }
+
+    /// Returns `true` if the poll state is [`Done`].
```
eholk, line 23: Should `Done` be `Written`? Or should `Written` be renamed `Done`?

src/utils/metadata.rs
```diff
@@ -0,0 +1,105 @@
...
+    /// [`Taken`]: PollState::Taken
+    #[must_use]
+    pub(crate) fn is_taken(&self) -> bool {
+        matches!(self, Self::Taken)
+    }
+}
+
+#[derive(Debug)]
+pub(crate) struct Metadata {
+    index: usize,
```
eholk, line 42: Is the `index` needed or can you always recover it from the location in the `Vec` where the `Metadata` is stored?

src/utils/pin.rs
```diff
@@ -11,16 +11,15 @@ pub(crate) fn iter_pin_mut<T>(slice: Pin<&mut [T]>) -> impl Iterator<Item = Pin<
...
-// pub(crate) fn pin_project_array<T, const N: usize>(slice: Pin<&mut [T; N]>) -> [Pin<&mut T>; N] {
-//     // SAFETY: `std` _could_ make this unsound if it were to decide Pin's
-//     // invariants aren't required to transmit through arrays. Otherwise this has
-//     // the same safety as a normal field pin projection.
-//     unsafe { slice.get_unchecked_mut() }
-//         .each_mut()
-//         .map(|t| unsafe { Pin::new_unchecked(t) })
-// }
+// From: `futures_rs::join_all!` -- https://github.com/rust-lang/futures-rs/blob/b48eb2e9a9485ef7388edc2f177094a27e08e28b/futures-util/src/future/join_all.rs#L18-L23
+pub(crate) fn iter_pin_mut_vec<T>(slice: Pin<&mut Vec<T>>) -> impl Iterator<Item = Pin<&mut T>> {
```
eholk, line 15: So I'm pretty sure this function is safe, but it feels a little off to me. The part that makes me nervous is that having a `&mut Vec<T>` means you could push a new item onto the Vec which could cause it to reallocate and move all of the elements in the container. But I don't think we can rely on `Vec<T>` being `Unpin`, which I think means we can't get into trouble without using something unsafe like `get_unchecked_mut`. So as long as we're careful in this function to not do anything like `push`, then I don't think there's any trouble.

Anyway, I don't think there's anything to do from this comment. This function just strikes me as the most subtle part of the PR so I wanted to add some comments about how I convinced myself it's correct.

eholk (approved):

src/future/join/vec.rs
yoshuawuyts: Unfortunately we can't: We have a `Vec<MaybeUninit<T>>`, not a `MaybeUninit<Vec<T>>`. We need to mark each entry as initialized in-place, which I don't think we can do without moving.

Transmuting seems like the better option in this case tbh. Miri isn't complaining about it, so I assume it should be fine?

src/utils/metadata.rs
yoshuawuyts: Oh yeah true, removed!

src/utils/metadata.rs
yoshuawuyts: Renamed the states to `Pending`, `Done`, and `Consumed` to mirror the `Poll` states + an extra "data was taken" state.

src/future/join/vec.rs
yoshuawuyts: good call!

src/future/join/vec.rs
yoshuawuyts: oh that's more robust in the face of faulty futures too. Good call!

src/future/join/vec.rs
yoshuawuyts: Good shout! - it turns out by removing the indexes from the metadata we can even turn this into `vec![Metadata::default(); futures.len()]`. Nice one!

src/utils/pin.rs
yoshuawuyts: Yeahhhhhh, I opened a Zulip issue about this, since it really does feel yank. Like you're saying: our usage of this is alright, but the abstraction: probably less so. It'd be nice if we could find something better, even if just for internal use this is fine for now.

PR #46, opened 2022-11-09
Init bench

### LINKED ISSUE #39: Author baseline benchmark
For #21 we're planning to implement what's essentially an optimization. We shouldn't do anything without at least having some form of benchmark, so we should start there.

Maybe we can borrow something from the futures-heavy rustc perf suite?

yoshuawuyts:
This initializes the benchmarks, starting with `stream::Merge`. Closes #39. cc/ @eholk

Initially `impl Merge for Vec` turned out to be so slow, the benchmark wouldn't even complete. We had to eliminate the fairness to make it work. This is what it's like now without fairness, but with allocations for the indexes:

merge-10-Criterion-rs

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,9 +13,14 @@ authors = [
     "Yoshua Wuyts <yoshuawuyts@gmail.com>"
 ]
 
+[[bench]]
+name = "bench"
+harness = false
+
 [dependencies]
 futures-core = "0.3"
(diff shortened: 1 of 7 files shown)


yoshuawuyts:
Removed the intermediate index allocations, still no fairness, but seeing a 75% performance improvement:

merge-10-Criterion-rs (1)

PR #48, opened 2022-11-09
Add optimized datastructure for tracking PollStates

### LINKED ISSUE #42: Inline `PollState` in future when possible
`JoinState` is 2-bits of state per future, which means on a u64 we should be able to inline the state for `Vec<impl Future>` of up to 32 items. We suspect that should be the majority of cases.

That would bring `vec::Join` down from 2 allocations (metadata + output) to just a single allocation (output) in the majority of cases.

michaelwoerister:
Closes #42. Not very polished yet but it should get the job done.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,4 +24,4 @@ pin-project = "1.0.8"
 [dev-dependencies]
 futures-lite = "1.12.0"
 futures = "0.3.25"
-criterion = "0.3"
+criterion = { version = "0.3", features = ["async", "async_futures"] }
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -1,6 +1,7 @@
 criterion::criterion_main!(merge::merge_benches, join::join_benches, race::race_benches);
 
 mod merge {
+    use criterion::async_executor::FuturesExecutor;
     use criterion::{black_box, criterion_group, Criterion};
     use futures_concurrency::prelude::*;
     use futures_lite::future::block_on;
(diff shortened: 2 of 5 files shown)


yoshuawuyts (changes requested):
`miri` seems to be hanging on CI. I'm not sure what's causing that tbh, but we should probably resolve it before merging.

michaelwoerister:
Miri probably is just timing out because the test is doing too much work. Let's see if reducing the number of iterations for miri solves the problem.

yoshuawuyts (approved):
The code looks good, and I'm happy to merge this as is. For performance though, there is a bit of question to what's going on. I've tested this against the suite from: https://github.com/yoshuawuyts/futures-concurrency/pull/52

## Benchmarks

### 10 items

We're seeing roughly similar performance here, but with less deviation which seems promising:

vec-join-10-Criterion-rs


### 100 items

This seems like a slight performance slowdown; the 1000 item case too.

vec-join-100-Criterion-rs

michaelwoerister:
Slower? We can't have that! 😉

Is there a convenient way to compare to git branches with criterion?

michaelwoerister:
This version should do a bit better, but it still seems to be slower. I'm not sure if this optimization is worth the trouble.

yoshuawuyts:
Yeah the benches I'm now getting are:

```text
❯ critcmp main patch -f vec::join
vec::join 10      1.00    646.7±5.73ns        ? ?/sec    1.01   654.0±34.59ns        ? ?/sec
vec::join 100     1.00      3.3±0.17µs        ? ?/sec    1.18      3.9±0.12µs        ? ?/sec
vec::join 1000    1.00     27.3±0.21µs        ? ?/sec    1.20     32.8±1.55µs        ? ?/sec
```

## Diagnosis

But I think I may have found something? The main idea here is that if we have a number of entries which is less than the size of a pointer, we can inline the entries in the space of the pointer. However, if we validate the sizes, it turns out we're not quite doing that right now!

### Test
```rust
#[test]
fn check_bit_sizes() {
    const BYTE: usize = 8;
    assert_eq!(std::mem::size_of::<EntriesWord>() * BYTE, 64);
    assert_eq!(std::mem::size_of::<PollStateEntries>() * BYTE, 64);
}
```
### Output
```text
running 1 test
thread 'utils::poll_state::check_bit_sizes' panicked at 'assertion failed: `(left == right)`
  left: `192`,
 right: `64`', src\utils\poll_state.rs:322:5
```

It seems the `PollStateEntries` enum is not 1 byte, but 3 bytes. I suspect that may be messing with cache coherence, and be the cause of the performance regression in the vec case.

## Potential Directions

```rust
/// The max number of entries `PollStates` can store without heap allocations.
const MAX_INLINE_ENTRY_COUNT: usize = ENTRIES_PER_WORD;
const ENTRIES_PER_WORD: usize = ((std::mem::size_of::<EntriesWord>() * 8) / 2);

enum PollStateEntries {
    Inline(EntriesWord),
    Boxed(Vec<EntriesWord>),
}
```

A few things stand out to me here:
- `EntriesWord` is a u64, but `Vec<EntriesWord>` is a `usize`. These may be the same on 64-bit machines, but they aren't necessarily.
- `ENTRIES_PER_WORD` will compile down to `32` on 64-bit machines. If you put that in an enum however, you need another byte to allocate the label on.

What I'm wondering is if by changing these variables around we can get the size of `PollStateEntries` down to 1 byte. Some combination of using `usize` everywhere and leaving space for the entries in the label. Or switching `ENTRIES_PER_WORD` to be `NonZeroUsize` might be able to help. I'm not sure, but it might be worth playing around with?

yoshuawuyts:
Oh actually: we're a bit off here - the assumption that "pointer to vec is 64 bits" is wrong. A `Vec` is a wide pointer, containing both the actual pointer to the heap, but also the length. That means it's three words wide:

```rust
assert_eq!(std::mem::size_of::<Vec<()>>() * BYTE, 192);
```

The issue _isn't_ that we're not aligning our usizes and what not right. There's plenty of space for that. In fact, if we do it right we should be able to store up to 3 words worth of entries in the space of a single `Vec`. So I'm not really sure what else might be causing it. It really might just be some sort of alignment thing?

yoshuawuyts:
Okay, I did find another interesting bit! I was looking at the size of `PollStateEntries`, but I should have been looking at the size of `PollStates` itself. It's defined as:

```rust
pub(crate) struct PollStates {
    len: usize,
    entries: PollStateEntries,
}
```

What we'd want is for `PollStates` to be the same size as `Vec<PollState>`. Instead, it's actually larger!

```rust
assert_eq!(std::mem::size_of::<Vec<PollState>>() * BYTE, 192); // ✅ 192 bits
assert_eq!(std::mem::size_of::<PollStates>() * BYTE, 192);     // ❌ 256 bits
```

And if you think about it: this makes sense. In the enum we're making space for `Vec<PollState>` which keeps a pointer + length. But we then wrap that in another struct which _also_ keeps track of length. That means we're tracking the length twice. And if you add on top of that what is potentially an `EntriesWord` which doesn't account for the enum label sizes - we're now larger than the original `Vec<PollState>`.

`PollStates::len` should only be needed for the `PollStateEntries::inline` case. I wonder if we put it in the enum's arm whether that can reduce the total size of the struct. And in turn make the `Boxed` case faster, since it will not need to track `len` in two places at the same time.

There might be other optimizations possible too, wrt accounting for the labels and what not. But that seems like the right place to start perhaps?

michaelwoerister:
The length of `Vec<EntriesWord>` is not the same as the number of entries, since the last word might not be fully used.

michaelwoerister:
Closing in favor of the much simpler https://github.com/yoshuawuyts/futures-concurrency/pull/78.

PR #51, opened 2022-11-10
reuse inline wakers

yoshuawuyts:
This reuses the waker instances, which should greatly reduce allocations.
# Benchmarks

These benchmarks are compared to the baseline, where we didn't provide inline wakers. It regresses somewhat, but not nearly as much as shown in https://github.com/yoshuawuyts/futures-concurrency/pull/50#issue-1442566344. Furthermore it makes wake times far more _consistent_, which gives me hope that this will become easier to profile, to optimize, and with some moderate effort we'll be able to bring the performance numbers down to where they were before our introduction of this.

## merge 10
merge-10-Criterion-rs (2)

## merge 100

merge-100-Criterion-rs

## merge 1000

merge-1000-Criterion-rs

### DIFF
--- a/src/stream/merge/vec.rs
+++ b/src/stream/merge/vec.rs
@@ -1,12 +1,12 @@
 use super::Merge as MergeTrait;
 use crate::stream::IntoStream;
-use crate::utils::{self, Fuse, RandomGenerator};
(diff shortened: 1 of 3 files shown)


yoshuawuyts:
Thinking about it more: this degree of regressions seems acceptable to me. With this we should have a solid foundation to build on, benchmark, and optimize further.

PR #57, opened 2022-11-11
Implement "perfect" waking for `impl Merge for Vec` (2/2)

yoshuawuyts:
Continues the work started in https://github.com/yoshuawuyts/futures-concurrency/pull/50, removing the final set of allocations. The strategy used for this is to create an initial `WakerList`, which we populate with intermediate wakers on the first call to `poll`. In a future patch we may want to create a static variant of this as well for use in our fixed-length combinators.

I also took the opportunity to rename `StreamWaker` to `InlineWaker`, since we may want to use this mechanism for the other combinators too.

## Benchmarks

Performance seems to be about 1.5-2x as good as before, which is expected since we no longer allocate on each call to `poll`:
```text
❯ critcmp main patch -f vec::merge
vec::merge 10      1.89      5.3±0.15µs        ? ?/sec    1.00      2.8±0.08µs        ? ?/sec        
vec::merge 100     1.68     75.3±0.56µs        ? ?/sec    1.00     44.8±1.11µs        ? ?/sec        
vec::merge 1000    1.46      2.8±0.02ms        ? ?/sec    1.00  1946.8±40.07µs        ? ?/sec   
```


src/stream/merge/vec.rs
```diff
@@ -66,12 +58,16 @@ where
     fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
         let mut this = self.project();
 
+        if !this.wakers.has_parent() {
+            this.wakers.set_parent(cx.waker());
```
eholk, line 62: I think this will cause problems if the waker changes in between calls to `poll_next`. As far as I know, we aren't guaranteed to have the same context/waker in each call to `poll`. For example, the merge stream could be sent to a different task/thread or moved into a different container. We could end up with a problem where we try to wake using the old waker rather than the new one. (It'd probably be good to have test coverage for this scenario, but that probably makes sense to add as a later PR.)

What I'd suggest is having `WakerList` have a `parent: Arc<Mutex<Option<Waker>>>` field and have each of the inner wakers have a clone of the same thing. Then at each call to `poll_next`, you can change the waker that's stored in the parent slot and each of the `InnerWaker`s will automatically access the most recent copy.

eholk: Oh, or another option is to put an `Option<Waker>` field in the `Readiness` set and then you don't need an additional `Arc` and `Mutex`.

eholk (changes requested):
Overall this looks really good, and the 2x speedup is awesome!

I think there's a problem where if the waker changes between calls to `poll_next` we'll still end up waking using the old waker, so it'd be good to make sure that's not an issue first :)

src/stream/merge/vec.rs
yoshuawuyts: Great call, fixed!

yoshuawuyts:
Updated with feedback from @eholk, we now set the parent waker on `readiness` at the start of each `poll` invocation. Because we already took out a lock on `readiness`, that effectively came at no cost. Meaning that compared to our baseline, we keep very similar performance improvements:

```text
❯ critcmp main patch  -f vec::merge
vec::merge 10      1.94      5.3±0.58µs        ? ?/sec    1.00      2.7±0.11µs        ? ?/sec
vec::merge 100     1.47     68.5±7.33µs        ? ?/sec    1.00     46.5±1.75µs        ? ?/sec
vec::merge 1000    1.13      2.2±0.09ms        ? ?/sec    1.00  1934.4±40.34µs        ? ?/sec
```

yoshuawuyts:
Just ran this against the pre-optimization baseline, and with this we're ranging between 2-20x as fast compared to our baseline from a week ago:

```text
❯ critcmp baseline patch  -f vec::merge
group              baseline                               patch
-----              --------                               -----
vec::merge 10      1.76      4.9±0.31µs        ? ?/sec    1.00      2.8±0.03µs        ? ?/sec
vec::merge 100     8.00    360.6±3.84µs        ? ?/sec    1.00     45.1±0.79µs        ? ?/sec
vec::merge 1000    19.27    38.0±0.41ms        ? ?/sec    1.00  1971.8±60.36µs        ? ?/sec
```

PR #58, opened 2022-11-12
Make implementation of `Race` for tuple fair

### LINKED ISSUE #44: Make `impl Race for tuple` fair
Right now we poll it in linear order. Just like with `impl Merge for tuple` we should make it so the poll order is randomized to preserve fairness.

matheus-consoli:
Closes #44


yoshuawuyts (approved):
This looks great!

PR #59, opened 2022-11-12
Remove duplication of `gen_conditions`

matheus-consoli:
This removes the macro duplication from #58.

We add two new parameters to the macro: the method to be called (either `poll` or `poll_next`), and the match arms.

One nice thing is that by receiving the match arms, we can now reference some things that before were passed as arguments (such as `pending`).


yoshuawuyts (approved):
This looks great; thank you so much!

PR #60, opened 2022-11-12
Use `array::from_fn` to safely create output array of `MaybeUninit`

matheus-consoli:
There're no performance impacts, just remove some unsafe using `array::from_fn`

### DIFF
--- a/src/future/join/array.rs
+++ b/src/future/join/array.rs
@@ -53,14 +53,12 @@ where
         }
 
         if all_done {
+            use core::array;
             use core::mem::MaybeUninit;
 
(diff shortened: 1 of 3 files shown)


matheus-consoli:
oh, wait, am I wrong or does this leaks because we're not calling `assume_unit()` anymore?

yoshuawuyts (approved):
This looks good! - For future reference: we're running the tests through `miri`, which means that if this leaked it should be caught. As part of #22 we may restructure this once more; but for now this looks like a pretty straightforward improvement on what we have!

PR #71, opened 2022-11-14
make `{array,vec}::race` fair

yoshuawuyts:
Ref #22 

Makes `{array,vec}::race` fair. Apparently we did not yet implement fair scheduling for the `race` combinators; this fixes it.

Additionally it seems we don't have a need for `PollState` in these structs because on the first "success" poll we immediately terminate. So we don't need to track state over time.

There is some perf regression, but that's expected because we do more work. But also: I'm not sure how good our `race` benches actually are since they don't actually measure repeated polling - they will always return on the first call to `poll`. For it to really show we'd have to author benches which return on the `nth` poll - but that's significantly harder and probably not worth it.

### DIFF
--- a/src/future/race/array.rs
+++ b/src/future/race/array.rs
@@ -1,3 +1,5 @@
+use crate::utils::{self, RandomGenerator};
+
 use super::Race as RaceTrait;
 
 use core::fmt;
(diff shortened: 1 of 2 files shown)


yoshuawuyts:
This should be pretty straightforward; going to be merging a few of these.

PR #72, opened 2022-11-14
Remove `MaybeDone` for `array::join`

yoshuawuyts:
Ref #22. Moves us away from `MaybeDone` for `array::join`.

## Performance

Performance appears to remain about equal as it was before, with a performance diff which can easily be attributed to noise:

```text
❯ critcmp main patch  -f array::join
group               main                                   patch
-----               ----                                   -----
array::join 10      1.03   589.1±26.71ns        ? ?/sec    1.00   571.7±12.23ns        ? ?/sec
array::join 100     1.00      3.7±0.03µs        ? ?/sec    1.05      3.8±0.16µs        ? ?/sec
array::join 1000    1.00     33.0±0.34µs        ? ?/sec    1.03     33.8±1.01µs        ? ?/sec
```

### DIFF
--- a/src/future/join/array.rs
+++ b/src/future/join/array.rs
@@ -1,12 +1,15 @@
 use super::Join as JoinTrait;
-use crate::utils::MaybeDone;
+use crate::utils;
+use crate::utils::PollState;
 
+use core::array;
 use core::fmt;
 use core::future::{Future, IntoFuture};
+use core::mem::{self, MaybeUninit};
 use core::pin::Pin;
 use core::task::{Context, Poll};
 
-use pin_project::pin_project;
+use pin_project::{pin_project, pinned_drop};
 
 /// Waits for two similarly-typed futures to complete.
 ///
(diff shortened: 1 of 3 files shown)


src/future/join/array.rs
```diff
@@ -16,12 +19,44 @@ use pin_project::pin_project;
...
+}
+
+impl<Fut, const N: usize> JoinTrait for [Fut; N]
+where
+    Fut: IntoFuture,
+{
+    type Output = [Fut::Output; N];
+    type Future = Join<Fut::IntoFuture, N>;
+
+    fn join(self) -> Self::Future {
```
poliorcetics, line 57: I wonder if marking this with `#[inline]` would have positive effects on perf accross crate boundaries

poliorcetics: (I'm launching this without even knowing if it could be part of a perf-limited code path, that's not an expert opinion, I'm jsut exploring the open PRs)

Issue #21, opened 2022-11-04
Keep track of woken futures


yoshuawuyts:
Instead of each `poll` call being `O(N)` in time, it should be `O(1)` in time, by tracking which futures have been woken on each iteration, and only waking those.

This acts as a guard against faulty manually implemented futures which don't have a fast return path in the case the wake is a no-op. https://github.com/yoshuawuyts/futures-concurrency/pull/8 is a first attempt at such a patch, but needs more work to finish up.

PR #74, opened 2022-11-14
Remove `MaybeDone` from `tuple::join`

matheus-consoli:
Ref #22 

Remove the need for `MaybeDone` on `tuple::join` by creating for each future three fields: the Future itself; its result; its state.

I had to add a dependency on `paste` to create the `$F_fields`.

Bench:
```text
>> critcmp main patch -f tuple::join
tuple::join 10    1.00    232.6±9.13ns        ? ?/sec    1.14    266.0±3.09ns        ? ?/sec
```

### DIFF
--- a/src/future/join/tuple.rs
+++ b/src/future/join/tuple.rs
@@ -1,5 +1,5 @@
 use super::Join as JoinTrait;
-use crate::utils::PollState;
+use crate::utils::{self, PollState, PollStates};
 
 use core::fmt::{self, Debug};
 use core::future::{Future, IntoFuture};
@@ -7,31 +7,95 @@ use core::mem::MaybeUninit;
 use core::pin::Pin;
 use core::task::{Context, Poll};
 
-use pin_project::{pin_project, pinned_drop};
+use pin_project::pin_project;
 
-use paste::paste;
+macro_rules! maybe_poll {
+    ($idx:tt, $len:ident, $this:ident, $fut:ident, $cx:ident) => {
+        if $this.states[$idx].is_pending() {
+            if let Poll::Ready(out) = $this.$fut.poll($cx) {
+                $this.outputs.$idx = MaybeUninit::new(out);
+                $this.states[$idx] = PollState::Done;
+                *$this.len -= 1;
+            }
+        }
+    };
+}
+
+macro_rules! poll_all_pending {
+    (@inner 0, $len:ident, $this:ident, $cx:ident, ($fut:ident, $($rest:ident,)*)) => {
+        maybe_poll!(0, $len, $this, $fut, $cx);
+        poll_all_pending!(@inner 1, $len, $this, $cx, ($($rest,)*));
+    };
+    (@inner 1, $len:ident, $this:ident, $cx:ident, ($fut:ident, $($rest:ident,)*)) => {
+        maybe_poll!(1, $len, $this, $fut, $cx);
+        poll_all_pending!(@inner 2, $len, $this, $cx, ($($rest,)*));
+    };
+    (@inner 2, $len:ident, $this:ident, $cx:ident, ($fut:ident, $($rest:ident,)*)) => {
+        maybe_poll!(2, $len, $this, $fut, $cx);
+        poll_all_pending!(@inner 3, $len, $this, $cx, ($($rest,)*));
+    };
+    (@inner 3, $len:ident, $this:ident, $cx:ident, ($fut:ident, $($rest:ident,)*)) => {
+        maybe_poll!(3, $len, $this, $fut, $cx);
+        poll_all_pending!(@inner 4, $len, $this, $cx, ($($rest,)*));
+    };
+    (@inner 4, $len:ident, $this:ident, $cx:ident, ($fut:ident, $($rest:ident,)*)) => {
+        maybe_poll!(4, $len, $this, $fut, $cx);
+        poll_all_pending!(@inner 5, $len, $this, $cx, ($($rest,)*));
+    };
+    (@inner 5, $len:ident, $this:ident, $cx:ident, ($fut:ident, $($rest:ident,)*)) => {
+        maybe_poll!(5, $len, $this, $fut, $cx);
+        poll_all_pending!(@inner 6, $len, $this, $cx, ($($rest,)*));
+    };
+    (@inner 6, $len:ident, $this:ident, $cx:ident, ($fut:ident, $($rest:ident,)*)) => {
+        maybe_poll!(6, $len, $this, $fut, $cx);
+        poll_all_pending!(@inner 7, $len, $this, $cx, ($($rest,)*));
+    };
+    (@inner 7, $len:ident, $this:ident, $cx:ident, ($fut:ident, $($rest:ident,)*)) => {
+        maybe_poll!(7, $len, $this, $fut, $cx);
+        poll_all_pending!(@inner 8, $len, $this, $cx, ($($rest,)*));
+    };
+    (@inner 8, $len:ident, $this:ident, $cx:ident, ($fut:ident, $($rest:ident,)*)) => {
+        maybe_poll!(8, $len, $this, $fut, $cx);
+        poll_all_pending!(@inner 9, $len, $this, $cx, ($($rest,)*));
+    };
+    (@inner 9, $len:ident, $this:ident, $cx:ident, ($fut:ident, $($rest:ident,)*)) => {
+        maybe_poll!(9, $len, $this, $fut, $cx);
+        poll_all_pending!(@inner 10, $len, $this, $cx, ($($rest,)*));
+    };
+    (@inner 10, $len:ident, $this:ident, $cx:ident, ($fut:ident, $($rest:ident,)*)) => {
+        maybe_poll!(10, $len, $this, $fut, $cx);
+        poll_all_pending!(@inner 11, $len, $this, $cx, ($($rest,)*));
+    };
+    (@inner 11, $len:ident, $this:ident, $cx:ident, ($fut:ident, $($rest:ident,)*)) => {
+        maybe_poll!(11, $len, $this, $fut, $cx);
+        poll_all_pending!(@inner 12, $len, $this, $cx, ($($rest,)*));
+    };
+    (@inner 12, $len:ident, $this:ident, $cx:ident, ($fut:ident, $($rest:ident,)*)) => {
+        maybe_poll!(12, $len, $this, $fut, $cx);
+    };
+    (@inner $ignore:literal, $len:ident, $this:ident, $cx:ident, ()) => { };
+    ($len:ident, $this:ident, $cx:ident, $($F:ident,)*) => {
+        poll_all_pending!(@inner 0, $len, $this, $cx, ($($F,)*));
+    };
+}
 
 macro_rules! impl_join_tuple {
     ($StructName:ident $($F:ident)*) => {
-        paste!{
-            /// Waits for two similarly-typed futures to complete.
-            ///
-            /// This `struct` is created by the [`join`] method on the [`Join`] trait. See
-            /// its documentation for more.
-            ///
-            /// [`join`]: crate::future::Join::join
-            /// [`Join`]: crate::future::Join
-            #[pin_project(PinnedDrop)]
-            #[must_use = "futures do nothing unless you `.await` or poll them"]
-            #[allow(non_snake_case)]
-            pub struct $StructName<$($F: Future),*> {
-                done: bool,
-                $(
-                    #[pin] $F: $F,
-                    [<$F _out>]: MaybeUninit<$F::Output>,
-                    [<$F _state>]: PollState,
-                )*
-            }
+        /// Waits for two similarly-typed futures to complete.
+        ///
+        /// This `struct` is created by the [`join`] method on the [`Join`] trait. See
+        /// its documentation for more.
+        ///
+        /// [`join`]: crate::future::Join::join
+        /// [`Join`]: crate::future::Join
+        #[pin_project]
+        #[must_use = "futures do nothing unless you `.await` or poll them"]
+        #[allow(non_snake_case)]
+        pub struct $StructName<$($F: Future),*> {
+            len: u32,
+            $(#[pin] $F: $F,)*
+            outputs: ($(MaybeUninit<$F::Output>,)*),
+            states: PollStates,
         }
 
         impl<$($F),*> Debug for $StructName<$($F),*>
@@ -40,11 +104,10 @@ macro_rules! impl_join_tuple {
             $F::Output: Debug,
         )* {
             fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-                paste!{
-                    f.debug_tuple("Join")
-                        $(.field(&self.$F).field(&self.[<$F _state >]))*
-                        .finish()
-                }
+                f.debug_tuple("Join")
+                    .field(&($(&self.$F,)*))
+                    .field(&self.states)
+                    .finish()
             }
         }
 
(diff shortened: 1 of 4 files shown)


yoshuawuyts (approved):
This looks great, thank you so much!

src/future/join/tuple.rs
```diff
@@ -1,28 +1,37 @@
...
+            #[pin_project(PinnedDrop)]
+            #[must_use = "futures do nothing unless you `.await` or poll them"]
+            #[allow(non_snake_case)]
+            pub struct $StructName<$($F: Future),*> {
+                done: bool,
+                $(
+                    #[pin] $F: $F,
+                    [<$F _out>]: MaybeUninit<$F::Output>,
+                    [<$F _state>]: PollState,
+                )*
```
yoshuawuyts, line 33: We may want to create named fields here instead:

```suggestion
                futures: ($(#[pin]$F,)*),
                output: ($(MaybeUninit<$F::Output>,)*),
                state: ($(PollState,)*),
```

src/future/join/tuple.rs
```diff
@@ -50,11 +61,24 @@ macro_rules! impl_merge_tuple {
...
+                        all_done &= this.[<$F _state>].is_done();
+                    )*
+                }
 
                 if all_done {
                     *this.done = true;
-                    Poll::Ready(($(this.$F.take().unwrap()),*))
+                    paste! {
+                        Poll::Ready(($( unsafe { this.[<$F _out>].assume_init_read() }),*))
+                    }
```
yoshuawuyts, line 81: The regression we're seeing is probably because of this. The main performance benefit of #22 is that we prevent an extra copy at the end. Say we have a tuple `(A, B)`, the output would always be `(A::Output, B::Output)` unless the future is cancelled.

Because we know this we can then store a field in-line of: `(MaybeUninit<A::Output>, MaybeUninit<B::Output>)`, which once we know all fields have been initialized we can just return from the future as-is. This patch unfortunately doesn't quite do that; in order to mark the fields as "initialized" here, it does a move.

That's why I'm suggesting we instead find a way to store `futures`, `output`, and `state` as separate fields of tuples - and find a way to index into them.

src/future/join/tuple.rs
```diff
@@ -50,11 +61,24 @@ macro_rules! impl_merge_tuple {
...
+                // Poll futures
+                paste! {
+                    $(
+                        if this.[<$F _state>].is_pending() {
+                            if let Poll::Ready(out) = this.$F.poll(cx) {
+                                *this.[<$F _out>] = MaybeUninit::new(out);
+                                *this.[<$F _state>] = PollState::Done;
+                            }
+                        }
+                        all_done &= this.[<$F _state>].is_done();
```
yoshuawuyts, line 73: Oof haha, I forgot we had this. We should probably initialize a counter set to `len: utils::tuple_len!()` inside the struct, and then with each completed future count down. That will also set us up to enable https://github.com/yoshuawuyts/futures-concurrency/issues/21 to work, since we'll want to stop calling all wakers on each iteration anyway. But that doesn't have to be inside this patch.

yoshuawuyts (changes requested):
Actually, on second thought: I just noticed we have a perf regression and I think it might be because we're doing extra copies at the end. Meaning we've added extra machinery, but don't quite reap the benefits we hoped we would

I've added some in-line comments explaining what's going on. Do you think you could update it to work that way?

matheus-consoli:
I've reworked this a bit:
- no `paste` dep
- no tuple copy from `(MaybeUninit<A>,..)` to `(A,)`

but it got somewhat worst hahaha

I'll try to figure out a better way to impl this later, but let me know if you have any ideas :)

```text
>> critcmp main patch -f tuple::join
tuple::join 10    1.00    231.6±0.73ns        ? ?/sec    1.19    275.7±1.31ns        ? ?/sec
```

<details>
  <summary>for reference, the macro is desugaring `Join3<A, B, C>` as follows:</summary>

```rust
#[pin_project]
#[must_use = "futures do nothing unless you `.await` or poll them"]
#[allow(non_snake_case)]
pub struct Join3<A: Future, B: Future, C: Future> {
    len: u32,
    #[pin] A: A,
    #[pin] B: B,
    #[pin] C: C,
    outputs: (
        MaybeUninit<A::Output>,
        MaybeUninit<B::Output>,
        MaybeUninit<C::Output>,
    ),
    states: PollStates,
}

impl<A: Future, B: Future, C: Future> Future for Join3<A, B, C> {
    type Output = (A::Output, B::Output, C::Output);
    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
        let mut this = self.project();
        if this.states[0].is_pending() {
            if let Poll::Ready(out) = this.A.poll(cx) {
                this.outputs.0 = MaybeUninit::new(out);
                this.states[0] = PollState::Done;
                *this.len -= 1;
            }
        };
        if this.states[1].is_pending() {
            if let Poll::Ready(out) = this.B.poll(cx) {
                this.outputs.1 = MaybeUninit::new(out);
                this.states[1] = PollState::Done;
                *this.len -= 1;
            }
        };
        if this.states[2].is_pending() {
            if let Poll::Ready(out) = this.C.poll(cx) {
                this.outputs.2 = MaybeUninit::new(out);
                this.states[2] = PollState::Done;
                *this.len -= 1;
            }
        };
        if *this.len <= 0 {
            let out = unsafe {
                (this.outputs as *const _ as *const (A::Output, B::Output, C::Output)).read()
            };
            Poll::Ready(out)
        } else {
            Poll::Pending
        }
    }
}
```
</details>

src/future/join/tuple.rs
```diff
@@ -1,14 +1,85 @@
...
+        poll_all_pending!(@inner 12, $len, $this, $cx, ($($rest,)*));
+    };
+    (@inner 12, $len:ident, $this:ident, $cx:ident, ($fut:ident, $($rest:ident,)*)) => {
+        maybe_poll!(12, $len, $this, $fut, $cx);
+    };
+    (@inner $ignore:literal, $len:ident, $this:ident, $cx:ident, ()) => { };
+    ($len:ident, $this:ident, $cx:ident, $($F:ident,)*) => {
+        poll_all_pending!(@inner 0, $len, $this, $cx, ($($F,)*));
+    };
+}
```
yoshuawuyts, line 80: Can you explain how this works? I'm not sure I completely follow, heh.

I was somewhat hoping we could create something like [`gen_conditions!`](https://github.com/yoshuawuyts/futures-concurrency/blob/main/src/utils/tuple.rs#L38-L56) which can be reused between implementations. Would that be possible?

src/future/join/tuple.rs
```diff
@@ -21,8 +92,10 @@ macro_rules! impl_merge_tuple {
         #[must_use = "futures do nothing unless you `.await` or poll them"]
         #[allow(non_snake_case)]
         pub struct $StructName<$($F: Future),*> {
-            done: bool,
-            $(#[pin] $F: MaybeDone<$F>,)*
+            len: u32,
+            $(#[pin] $F: $F,)*
```
yoshuawuyts, line 96: Would it be possible to instead do:

```suggestion
            tuple: ($(#[pin] $F: $F,)*),
```

That way we can move the tuple fields into here basically in-place.

src/future/join/tuple.rs
matheus-consoli: sure!

The idea is quite simple: to unroll a for-loop through all tuple elements.

But this is kinda a mess, because accessing tuples on macros is messy.

My first attempt was to create something akin to `gen_conditions`, a `for i in 0..LEN` calling a macro for each index accessing the tuple element. But we cannot `$tuple.$i` because `$i` is a variable, not a literal, and this would translate to `tuple.i`.

So I tried to inline/unroll the for.

The way this goes is by taking all type parameters the JoinN has, "poping" the first element (`$fut`), and generating the poll for that future (`maybe_poll`) - along the way, we have the "iteration index" as literal. Next, we call the next index with the rest of the generics. Once the list is empty, the macro will match the `(@inner $ignore:literal, $len:ident, $this:ident, $cx:ident, ())` and generate nothing, stopping the recursion.

I can elaborate on it better later, documenting it on code.

We can use `trace_macros` to see how this expand:
```rust
note: trace_macro
   --> src/future/join/tuple.rs:163:1
    |
163 | impl_join_tuple! { Join3 A B C }
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: expanding `poll_all_pending! { LEN, this, cx, A, B, C, }`
    = note: to `poll_all_pending! (@ inner 0, LEN, this, cx, (A, B, C,)) ;`
    = note: expanding `poll_all_pending! { @ inner 0, LEN, this, cx, (A, B, C,) }`
    = note: to `maybe_poll! (0, LEN, this, A, cx) ; poll_all_pending!
            (@ inner 1, LEN, this, cx, (B, C,)) ;`
    = note: expanding `maybe_poll! { 0, LEN, this, A, cx }`
    = note: to `if this.states [0].is_pending()
            {
                if let Poll :: Ready(out) = this.A.poll(cx)
                {
                    this.outputs.0 = MaybeUninit :: new(out) ; this.states [0] = PollState
                    :: Done ; * this.len -= 1 ;
                }
            }`
    = note: expanding `poll_all_pending! { @ inner 1, LEN, this, cx, (B, C,) }`
    = note: to `maybe_poll! (1, LEN, this, B, cx) ; poll_all_pending!
            (@ inner 2, LEN, this, cx, (C,)) ;`
    = note: expanding `maybe_poll! { 1, LEN, this, B, cx }`
    = note: to `if this.states [1].is_pending()
            {
                if let Poll :: Ready(out) = this.B.poll(cx)
                {
                    this.outputs.1 = MaybeUninit :: new(out) ; this.states [1] = PollState
                    :: Done ; * this.len -= 1 ;
                }
            }`
    = note: expanding `poll_all_pending! { @ inner 2, LEN, this, cx, (C,) }`
    = note: to `maybe_poll! (2, LEN, this, C, cx) ; poll_all_pending!
            (@ inner 3, LEN, this, cx, ()) ;`
    = note: expanding `maybe_poll! { 2, LEN, this, C, cx }`
    = note: to `if this.states [2].is_pending()
            {
                if let Poll :: Ready(out) = this.C.poll(cx)
                {
                    this.outputs.2 = MaybeUninit :: new(out) ; this.states [2] = PollState
                    :: Done ; * this.len -= 1 ;
                }
            }`
    = note: expanding `poll_all_pending! { @ inner 3, LEN, this, cx, () }`
    = note: to ``
```

src/future/join/tuple.rs
matheus-consoli: I'm currently trying to figure out how to do this!

`tuple: ($(#[pin] $F: $F,)*)` doesn't work because `pin_project` doesn't support pinning this way, like `(#[pin] T, #[pin] S)`

matheus-consoli: @yoshuawuyts, I got something!

It's on a second branch, [here](https://github.com/yoshuawuyts/futures-concurrency/compare/main...matheus-consoli:futures-concurrency:remove-maybedone-from-join-tuple-take2).

I took some inspiration from #84 to use the pinned futures inside a tuple(struct), and it kinda works.

We have some performance gain, but it's kinda negligible :disappointed: (especially because of the macro-heavy impl orientation)

```text
>> critcmp main patch -f tuple::join
tuple::join 10    1.03    239.3±1.06ns        ? ?/sec    1.00    231.3±0.79ns        ? ?/sec
```

yoshuawuyts: Oh yeah, I like the intermediate struct approach!

The reason why there's a small perf cost might be because you're using `ptr::read` there which [ends up doing an extra copy]((https://doc.rust-lang.org/src/core/ptr/mod.rs.html#1117))? Instead I think if it's at all possible it might be worth attempting to perform a direct transmute? This might require adding a `#[repr(transparent)]` to the `Futures` struct so that the layout is guaranteed to be the same as the tuples it stores?

src/future/join/tuple.rs
yoshuawuyts: I seee, that makes sense! Thank you for explaining. I think something which maybe threw me off a bit is the fact that the arm which matches on "tuples of length 12" is marked as `@inner 1`, and the arm which matches on "tuples of length 1" is marked as `@inner 11`. If the count could match the number of tuples it operates on it might be easier to reason about perhaps?

Now that you've explained it I'm on board with this approach! - If we could figure out a way to make this reusable that'd still be nice, but we don't have to do that in this patch.

src/future/join/tuple.rs
matheus-consoli: oh, I don't think `#[repr(transparent)] Futures` is possible, `transparent` only allows one field to have a non-zero size, the `Futures` have many.

matheus-consoli: I'm trying to `transmute`, but maybe I have found a limitation (?)

I can transmute from `(MaybeUninit<ConcreteTypeA>, MaybeUninit<ConcreteTypeB>)` to `(ConcreteTypeA, ConcreteTypeB)`, but I can't to the same from `(MaybeUninit<GenericTypeA>, MaybeUninit<GenericTypeB>)`, [play](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e948fe4a441734c81c5ce7b35290a23b)

I'll try again after lunch

yoshuawuyts: > oh, I don't think `#[repr(transparent)] Futures` is possible, `transparent` only allows one field to have a non-zero size, the `Futures` have many.

oof, TIL :')

matheus-consoli: oook, it seems to me that we cannot transmute `(MaybeUninit<T>, MaybeUninit<U>)` to `(T, U)`.

I'm not sure why, but I guess it's the same underlying problem from rust-lang/rust#61956, so casting the tuples appears to be the best option so far

src/future/join/tuple.rs
matheus-consoli: hm, but the "tuples of length 12" are not marked as `@inner 1`, nor the "tuples of length 1" are marked as `@inner 11`

it recursively unrolls to the size of the tuple, starting from 0 to the size - 1

assume we can read `@inner` as `@access_index_on_tuple` and we're unrolling a `(A, B, C)`

the macro goes as:
> `poll_all_pending!(tuple, (A, B, C))`
> `@access_index 0 on tuple $tuple remaining (A, B, C)`
> `$tuple.0` && `@access_index 1 on tuple $tuple remaining (B, C)`
> `$tuple.1` && `@access_index 2 on tuple $tuple remaining (C,)`
> `$tuple.2` && `@access_index 3 on tuple $tuple remaining ()`
> oh, remaining is empty, nothing left, stop here (aka `(@inner $ignore:literal, $len:ident, $this:ident, $cx:ident, ()) => { };
`

does that make sense?

yoshuawuyts: It does, thank you!

yoshuawuyts (approved):
I've just approved https://github.com/yoshuawuyts/futures-concurrency/pull/96. It seems this PR has merge conflicts anyway; if those can be resolved I think this should be good to go as well.

Either as part of this PR, or as part of a follow-up PR we should switch this implementation over to using [`tuple_for_each!`](https://github.com/yoshuawuyts/futures-concurrency/pull/96/files#diff-10eb736ee20e16b51fcc09b522c927cf43cbf8d4a5e55bfe40942b3718eddb64) as well, so the underlying implementation can be shared between all tuples.

src/future/join/tuple.rs
poliorcetics: Transmuting is not what you want here I think ?

```rust
let t = (MaybeUninit::<usize>::new(42), MaybeUninit::<u8>::new(12)); // source, can come from anywhere

/* ... */

// Where you wanted to transmute:
let (a, b) = t;
(a.assume_init(), b.assume_init())
```

This can be macro-ified using `$F` as the variable name: it won't even shadow the type names because that's two different namespaces (`let usize: usize = 4_usize;` is valid)

poliorcetics: Assuming the `t` is owned, this should not do any copy anywhere

matheus-consoli: ohh, thank you for pointing it out! I'll try that

src/future/join/tuple.rs
```diff
@@ -66,33 +138,36 @@ macro_rules! impl_merge_tuple {
...
             fn join(self) -> Self::Future {
                 let ($($F,)*): ($($F,)*) = self;
+                const LEN: u32 = utils::tuple_len!($($F,)*);
                 $StructName {
-                    done: false,
-                    $($F: MaybeDone::new($F.into_future())),*
+                    len: LEN,
+                    $($F: $F.into_future(),)*
+                    outputs: ($(MaybeUninit::<$F::Output>::uninit(),)*),
+                    states: PollStates::new(LEN as usize),
```
poliorcetics, line 151: PollArray is usable here

matheus-consoli:
I'm closing this PR in favor of #103 

I opted to start from a clean new branch instead of dealing with this branch conflicts :sweat_smile: 

Thank you all!

PR #79, opened 2022-11-15
inline `poll_states` and remove `Fuse` for `vec::merge`

yoshuawuyts:
Ref #22 

Counterpart to https://github.com/yoshuawuyts/futures-concurrency/pull/70. Removes the use of `Fuse` from `impl Merge for Vec`, and also inlines the `poll_states`.

## Performance

**Compared to main**
Performance remains roughly comparable.
```text
❯ critcmp main patch  -f vec::merge
vec::merge 10      1.04      2.7±0.12µs        ? ?/sec    1.00      2.6±0.05µs        ? ?/sec
vec::merge 100     1.00     43.2±1.93µs        ? ?/sec    1.10     47.5±0.97µs        ? ?/sec
vec::merge 1000    1.02      2.0±0.19ms        ? ?/sec    1.00      2.0±0.15ms        ? ?/sec
```

**Compared to last release**

The delta here remains mostly dominated by the move to "perfect" waker semantics:
```text
❯ critcmp patch baseline  -f vec::merge
group              baseline                               patch
-----              --------                               -----
vec::merge 10      1.93      4.9±0.31µs        ? ?/sec    1.00      2.6±0.05µs        ? ?/sec
vec::merge 100     7.60    360.6±3.84µs        ? ?/sec    1.00     47.5±0.97µs        ? ?/sec
vec::merge 1000    18.94    38.0±0.41ms        ? ?/sec    1.00      2.0±0.15ms        ? ?/sec
```

### DIFF
--- a/src/stream/merge/vec.rs
+++ b/src/stream/merge/vec.rs
@@ -1,6 +1,6 @@
 use super::Merge as MergeTrait;
 use crate::stream::IntoStream;
-use crate::utils::{self, Fuse, RandomGenerator, WakerList};
+use crate::utils::{self, PollState, PollStates, RandomGenerator, WakerList};
 
 use core::fmt;
 use futures_core::Stream;
@@ -20,10 +20,13 @@ where
     S: Stream,
 {
     #[pin]
-    streams: Vec<Fuse<S>>,
+    streams: Vec<S>,
     rng: RandomGenerator,
     complete: usize,
     wakers: WakerList,
+    state: PollStates,
+    done: bool,
+    len: usize,
 }
 
 impl<S> Merge<S>
@@ -31,11 +34,15 @@ where
     S: Stream,
 {
     pub(crate) fn new(streams: Vec<S>) -> Self {
+        let len = streams.len();
         Self {
-            wakers: WakerList::new(streams.len()),
-            streams: streams.into_iter().map(Fuse::new).collect(),
+            wakers: WakerList::new(len),
+            state: PollStates::new(len),
+            streams,
             rng: RandomGenerator::new(),
             complete: 0,
+            done: false,
+            len,
         }
     }
 }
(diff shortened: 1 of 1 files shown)


src/stream/merge/vec.rs
```diff
@@ -20,22 +20,29 @@ where
     S: Stream,
 {
     #[pin]
-    streams: Vec<Fuse<S>>,
+    streams: Vec<S>,
```
eholk, line 23: I wondered what the `Fuse` was for when I was looking at this code earlier, so I'm glad to see it going away and now I don't have to worry about it :)

src/stream/merge/vec.rs
```diff
@@ -20,22 +20,29 @@ where
...
     rng: RandomGenerator,
     complete: usize,
     wakers: WakerList,
+    state: PollStates,
+    done: bool,
+    len: usize,
```
eholk, line 29: I assume this is the number of streams we're merging? I might suggest `count` or `num_streams` as alternate names, but there's plenty of precedent for using `len` for this sort of thing, so I'm happy to leave it up to your judgement.

eholk: Actually, I'd suggest adding a `len()` method that returns `self.streams.len()` and that way we save 8 bytes off the size of the `Merge` struct.

src/stream/merge/vec.rs
```diff
@@ -92,6 +96,7 @@ where
                 }
                 Poll::Ready(None) => {
                     *this.complete += 1;
+                    this.state[index] = PollState::Consumed;
```
eholk, line 99: Since you have `.is_consumed()`, it might be good to have `.set_consumed()` also. That way you have some freedom to change the underlying representation without changing the API.

eholk (approved):
Looks good! I had a couple small suggestions for potential improvements, but I'll leave them up to you if you want to take them. Feel free to merge whenever!

src/stream/merge/vec.rs
yoshuawuyts: Oh I like that - might do that in a follow-up though since https://github.com/yoshuawuyts/futures-concurrency/pull/81 overhauls the `PollStates` system a bunch.

src/stream/merge/vec.rs
yoshuawuyts: Ah, I wish we could set methods! - Because we're dealing with a pin _projection_ it generates a struct for us containing the projection. Andy any `self` methods would need to point to that.

Because we don't control the projection we can't define methods. And so we basically have to define everything in-line. Which is like, another reason why I'd reallllly like us to have [safe pin projections](https://blog.yoshuawuyts.com/safe-pin-projections-through-view-types/), so we can split manually authored futures into separate internal methods etc. Not being able to do this is another reason authoring futures by hand is hard :/

PR #81, opened 2022-11-15
Provide const variants of the utils/ types

### LINKED ISSUE #76: Split `utils::WakerList` into `WakerVec` and `WakerArray`
This works towards #68, but we should have a const-variant of the existing `WakerList`. Probably one way to do it would be to separate it by name: e.g. `WakerVec` (dynamic size) vs `WakerArray` (static size). This would make PRs like https://github.com/yoshuawuyts/futures-concurrency/pull/75 faster and not rely on allocations.

### LINKED ISSUE #77: Rename `PollState::Done` to `PollState::Ready`
In streams it signals "has data available" rather than "has finished". Matching the `Poll` names feels like a much better fit than what we're currently doing.

yoshuawuyts:
Closes #76, Closes #77.

Creates non-allocating variants of the various types in `utils/`. Renames `PollState::Done` to `PollState::Ready`.

Once merged we should probably update all current PRs, since they touch `PollStates` in one way or another. Or perhaps better: we merge those PRs and then we rebase this PR to update the changes made there.


yoshuawuyts:
Rebased with all the changes in place. This is what the perf is like compared to the current `main` branch:

```rust
❯ critcmp main patch -f array
array::join 10    2.72  1546.7±45.93ns        ? ?/sec    1.00   568.9±17.86ns        ? ?/sec
array::join 100   1.13      4.0±0.08µs        ? ?/sec    1.00      3.5±0.08µs        ? ?/sec
array::join 1000  1.02     34.3±1.29µs        ? ?/sec    1.00     33.5±0.65µs        ? ?/sec
array::merge 10   1.00      2.9±0.05µs        ? ?/sec    1.09      3.2±0.05µs        ? ?/sec
array::merge 100  1.48     48.1±1.44µs        ? ?/sec    1.00     32.5±0.31µs        ? ?/sec
array::merge 1000 2.38  1877.5±57.35µs        ? ?/sec    1.00   790.2±15.13µs        ? ?/sec
array::race 10    1.14   694.7±18.30ns        ? ?/sec    1.00    608.2±8.40ns        ? ?/sec
array::race 100   1.10      4.9±0.20µs        ? ?/sec    1.00      4.4±0.16µs        ? ?/sec
array::race 1000  1.23     40.1±1.10µs        ? ?/sec    1.00     32.5±0.51µs        ? ?/sec
```

Overall these seem like pretty consistent performance improvements.

PR #84, opened 2022-11-15
feat: Remove `MaybeDone` for `tuple::join` (without additional deps)

poliorcetics:


poliorcetics:
I'm interested in the perf comparison with https://github.com/yoshuawuyts/futures-concurrency/pull/74 and the baseline :)

src/future/join/tuple.rs
```diff
@@ -1,16 +1,30 @@
...
+    ($mod_name:ident :: $StructName:ident $($F:ident)*) => {
+        mod $mod_name {
+            #[allow(unused_imports)] // Len == 0
+            use super::*;
+
+            pub(super) struct Outputs<$($F: Future),*> {
+                $(pub(super) $F: MaybeUninit<$F::Output>),*
+            }
+
+            pub(super) struct States {
```
matheus-consoli, line 22: you could use [`PollStates`](https://github.com/yoshuawuyts/futures-concurrency/blob/7f398ba5bab6416b0db3c0a93c080a23862e9f7e/src/utils/poll_state.rs#L67) here

I can run a comparison with the main, my branch, and ours after that :)

matheus-consoli: but here's the current results

```
>> critcmp main patch84 -f tuple::join
group             patch84                                   main
-----             -------                                   ----
tuple::join 10    1.00    213.1±0.96ns        ? ?/sec    1.12    239.5±0.56ns        ? ?/sec

```

yoshuawuyts: We have a `no_std` version of `PollStates` now: `utils::PollArray`.

yoshuawuyts:
I've just approved https://github.com/yoshuawuyts/futures-concurrency/pull/96 which builds on top of some of the concepts in this PR. I expect https://github.com/yoshuawuyts/futures-concurrency/pull/74 will be updated soon to match that as well. Even if we may not choose to merge this PR, the work you've done has been incredibly insightful, and I wanted to thank you for your contributions to the project! :pray:

poliorcetics:
Thanks ! Don't worry about this PR if #74 is better I have no problem merging it instead of this :)

PR #86, opened 2022-11-16
Make impl of Join for single element tuple return single elmnt tuple

matheus-consoli:
The `Join` impl for a single element tuple was returning the element itself, not a tuple - and I don't think it was intentional.

Just flipping a comma in the macro solves that.


yoshuawuyts (approved):
Oh great catch, thank you!

PR #92, opened 2022-11-16
Hide PollState variants

### LINKED ISSUE #88: Add `PollState{,s}::set_consumed`
Ref https://github.com/yoshuawuyts/futures-concurrency/pull/79#discussion_r1023319824; we should not only have a way to get poll states, but also to set them. Because #81 overhauls the system a bunch we should first merge that, or build directly on top of it.

matheus-consoli:
Closes #88 

I also "hid" `PollState::Ready` behind a `set_ready`. Now no one outside `utils/poll_state/poll_state.rs` references the variants by name.


yoshuawuyts (approved):
Thanks, this is great!

PR #93, opened 2022-11-17
Example: Happy eyeballs

eholk:
This adds an example program that demonstrates a happy eyeballs implementation, which is inspired by @njsmith's [talk](https://www.youtube.com/watch?v=oLkfnc_UMcE).

Joint work with @yoshuawuyts.

### DIFF
--- a/examples/happy_eyeballs.rs
+++ b/examples/happy_eyeballs.rs
@@ -34,7 +34,7 @@ async fn open_tcp_socket(
     attempts: u64,
 ) -> Result<TcpStream, Box<dyn Error + Send + Sync + 'static>> {
     let (mut sender, mut receiver) = oneshot::channel();
-    let mut futures = vec![];
+    let mut futures = Vec::with_capacity(attempts as usize);
 
     for attempt in 0..attempts {
         let tcp = TcpStream::connect((addr, port));
--- a/examples/happy_eyeballs.rs
+++ b/examples/happy_eyeballs.rs
@@ -1,4 +1,3 @@
-use async_std::io::prelude::*;
 use futures::channel::oneshot;
 use futures_concurrency::prelude::*;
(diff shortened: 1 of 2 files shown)


examples/happy_eyeballs.rs
```diff
@@ -0,0 +1,56 @@
...
+) -> Result<TcpStream, Box<dyn Error + Send + Sync + 'static>> {
+    let (init_send, init_recv) = oneshot::channel();
+    // Make the initial one start up right away
+    let _ = init_send.send(());
+    let mut futures = vec![];
+    let _last_receiver = (0..attempts).fold(init_recv, |event, attempt| {
+        let (send, recv) = oneshot::channel();
+        futures.push(async move {
+            // Wait for the previous one to complete
+            let _ = event.timeout(Duration::from_secs(attempt)).await?;
```
yoshuawuyts, line 39: This actually isn't quite right - in the case of a timeout here we don't start the next attempt, we appear to abort instead. The `?` here probably shouldn't be used.

yoshuawuyts:
Couldn't help myself; I've tinkered with this a bit more! Having a lot of fun with this too ^^
- I've removed some of the indentation, instead the code should read a bit more "flat".
- Instead of using a `fold` call which intereacted with `#[must_use]` and required `_` assignments, this now uses a flat `for`-loop updating the `(sender, receiver)` variables in a loop.
- I rewatched some of the talk, and it seems there wasn't a final timeout - which means we were doing more than actually needed as part of the core algorithm.
- Speaking of unnecessary things - since we're connecting to public domains anyway, I figured turning it into a quick `GET` request might be fun. So I added that ^^

yoshuawuyts:
We may want to do the `Ipv4`/`Ipv6` selection stuff. But that doesn't have to happen now. Merging, and we can always update that later!

PR #96, opened 2022-11-18
Implement "perfect" waking for `tuple::merge`

matheus-consoli:
This mimics the implementation for `array::merge`.

I used that hack I introduced on #74 in a more generalized form, we can rewrite the `gen_conditions` in terms of this new macro (basically the same way I do here)

Ref #21

### DIFF
--- a/src/stream/merge/tuple.rs
+++ b/src/stream/merge/tuple.rs
@@ -1,15 +1,41 @@
 use super::Merge as MergeTrait;
 use crate::stream::IntoStream;
-use crate::utils;
+use crate::utils::{self, PollArray, WakerArray};
 
 use core::fmt;
 use futures_core::Stream;
 use std::pin::Pin;
 use std::task::{Context, Poll};
 
-// TODO: handle none case
+macro_rules! poll_stream {
+    ($stream_idx:tt, $iteration:ident, $this:ident, $streams:ident, $cx:ident, $len_streams:ident) => {
+        if $stream_idx == $iteration {
+            match unsafe { Pin::new_unchecked(&mut $streams.$stream_idx) }.poll_next(&mut $cx) {
+                Poll::Ready(Some(item)) => {
+                    // Mark ourselves as ready again because we need to poll for the next item.
+                    $this
+                        .wakers
+                        .readiness()
+                        .lock()
+                        .unwrap()
+                        .set_ready($stream_idx);
+                    return Poll::Ready(Some(item));
+                }
+                Poll::Ready(None) => {
+                    *$this.completed += 1;
+                    $this.state[$stream_idx].set_consumed();
+                    if *$this.completed == $len_streams {
+                        return Poll::Ready(None);
+                    }
+                }
+                Poll::Pending => {}
+            }
+        }
+    };
+}
+
 macro_rules! impl_merge_tuple {
-    ($StructName:ident) => {
+    ($ignore:ident $StructName:ident) => {
         /// A stream that merges multiple streams into a single stream.
         ///
         /// This `struct` is created by the [`merge`] method on the [`Merge`] trait. See its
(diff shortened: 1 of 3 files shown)


matheus-consoli:
oh, I almost forgot the perf result

```text
>> critcmp main patch -f tuple::merge
tuple::merge 10    1.03    763.3±1.40ns        ? ?/sec    1.00    738.1±2.47ns        ? ?/sec
```

src/stream/merge/tuple.rs
```diff
@@ -84,33 +114,41 @@ macro_rules! impl_merge_tuple {
...
+                    }
+
+                    // unlock readiness so we don't deadlock when polling
+                    drop(readiness);
+
+                    // Obtain the intermediate waker.
+                    let mut cx = Context::from_waker(this.wakers.get(index).unwrap());
+
+                    // poll the `streams.{index}` stream
+                    utils::tuple_for_each!(poll_stream (index, this, streams, cx, LEN) $($F)*);
```
yoshuawuyts, line 145: Oh this is really nice!

src/stream/merge/tuple.rs
```diff
@@ -124,32 +162,36 @@ macro_rules! impl_merge_tuple {
...
+impl_merge_tuple! { merge3 Merge3  A B C }
+impl_merge_tuple! { merge4 Merge4  A B C D }
+impl_merge_tuple! { merge5 Merge5  A B C D E }
+impl_merge_tuple! { merge6 Merge6  A B C D E F }
+impl_merge_tuple! { merge7 Merge7  A B C D E F G }
+impl_merge_tuple! { merge8 Merge8  A B C D E F G H }
+impl_merge_tuple! { merge9 Merge9  A B C D E F G H I }
+impl_merge_tuple! { merge10 Merge10 A B C D E F G H I J }
+impl_merge_tuple! { merge11 Merge11 A B C D E F G H I J K }
+impl_merge_tuple! { merge12 Merge12 A B C D E F G H I J K L }
```
poliorcetics, line 189: A little nit: macro-highlighting is basically impossible to do well in Rust, but to simplify it for treesitter (the library used by github and lots of editors), we could make the input to this and other macros something like this:

`$mod_name: ident :: $StructName: ident <$($F: ident),+>`

That's not a blocker at all though, I just like the idea of using common Rust syntax so that macros are both easier to read and highlights.

Of course the issue here would be that `MergeX` is not really in the `mergeX` module, just wanted to get the idea out here :)

poliorcetics:
```rust
macro_rules! tuple_indexes {
    ($mod_name:ident :: $StructName:ident < $($F:ident),+ >) => {
        mod $mod_name {
            // Ensure indexes are in increasing order and start at 0 while being
            // very cheap to generate for the macro.
            //
            // If you want to ensure there are 256 elements or less, use repr(u8)
            #[repr(usize)]
            pub enum Indexes {
                $($F),+
            }

            pub const TUPLE_LEN: usize = match [$(Indexes::$F),+] {
                [.., last] => last as usize + 1,
            };
        }

        #[derive(Default)]
        pub struct $StructName {
            state: [bool; $mod_name::TUPLE_LEN],
        }

        impl $StructName {
            $(
                #[allow(non_snake_case)]
                pub fn $F(&self) -> (usize, bool) {
                    let index = tuple_indexes!(@index of $mod_name :: $F);
                    (index, self.state[index])
                }
            )+
        }
    };
    (@index of $mod_name: ident :: $F: ident) => {
        $mod_name::Indexes::$F as usize
    };
}

tuple_indexes!(test_1::Test1<A>);
tuple_indexes!(test_2 :: Test2 < A, B >);
tuple_indexes!(test_3 :: Test3 < A, B, C >);

#[test]
fn expected_indexes() {
    let t = Test1::default();
    assert_eq!(t.A().0, 0);

    let t = Test2::default();
    assert_eq!(t.A().0, 0);
    assert_eq!(t.B().0, 1);

    let t = Test3::default();
    assert_eq!(t.A().0, 0);
    assert_eq!(t.B().0, 1);
    assert_eq!(t.C().0, 2);
}
```

This could be used to simplify the macros, making rustc do less work ^^

matheus-consoli:
that's a very interesting approach, @poliorcetics!

I'll merge this PR as it is now, but I would like to see a scratch PR integrating your ideas - I think it would be very fair to see you in the contributors list!

poliorcetics:
> I'll merge this PR as it is now, but I would like to see a scratch PR integrating your ideas - I think it would be very fair to see you in the contributors list!

Done in #99 :) I'll try to run the benchmarks on my side to see if it helped

PR #100, opened 2022-11-19
Fix Race docs and cleanup unused items

### LINKED ISSUE #97: Fix docs
On https://docs.rs/futures-concurrency/7.0.0/futures_concurrency/future/trait.Race.html#tymethod.race the docs say:

> Awaits multiple futures simultaneously, returning the output of the futures once both complete.

We only wait for the first future to complete. We should fix that. First reported [here](https://www.reddit.com/r/rust/comments/yxoyme/comment/iwri7oc/).

poliorcetics:
- doc: small corrections and improvements
- chore: removed unused elements and imports

Closes #97


src/utils/poll_state/poll_state.rs
```diff
@@ -14,44 +14,34 @@ pub(crate) enum PollState {
...
     #[must_use]
     #[inline]
     pub(crate) fn is_ready(&self) -> bool {
         matches!(self, Self::Ready)
     }
 
-    /// Sets the poll state to [`Ready`]
-    ///
-    /// [`Ready`]: PollState::Ready
+    /// Sets the poll state to [`Ready`][Self::Ready].
```
yoshuawuyts, line 31: Can we keep the links in this file the way they were? This is the [style used by the stdlib](https://doc.rust-lang.org/src/core/iter/adapters/peekable.rs.html#4-11), and since we're aiming to PR this code into the stdlib I would like us to keep following the existing precedent.

yoshuawuyts (changes requested):
Just one issue, otherwise it looks great!

yoshuawuyts (approved):
On a second thought: this was on methods which aren't public, and will never be public. This is fine as-is!

PR #103, opened 2022-11-20
Remove `MaybeDone` from tuple::join

matheus-consoli:
Ref #22

Remove the need for `MaybeDone` on `Join` for tuples.

This is a spiritual successor of #74, thank you @yoshuawuyts and @poliorcetics for your reviews and ideas!

The perf results are really nice now:
```text
>> critcmp main maybegone -f tuple::join
group             main                                   maybegone
-----             ----                                   ---------
tuple::join 10    1.00    246.2±0.79ns        ? ?/sec    2.86    703.8±1.49ns        ? ?/sec
```

edit: removes reference to #21, as we're no longing implementing "perfect" waking

### DIFF
--- a/src/future/join/tuple.rs
+++ b/src/future/join/tuple.rs
@@ -1,5 +1,5 @@
 use super::Join as JoinTrait;
-use crate::utils::{PollArray, RandomGenerator, WakerArray};
+use crate::utils::PollArray;
 
 use core::fmt::{self, Debug};
 use core::future::{Future, IntoFuture};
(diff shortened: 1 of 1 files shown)


matheus-consoli:
oh, yes, this also makes the implementation of `join` fair

poliorcetics:
I don't understand the perf result, isn't `maybegone` slower ?

matheus-consoli:
ohnoooo, I'm stupid, I read the result backwards again 
this sucks, sorry

yoshuawuyts:
Yeah, that's a pretty big slowdown. To be honest: the rng stuff probably shouldn't be included since all futures only poll once - and then have to wait for each other anyway.

Given that https://github.com/yoshuawuyts/futures-concurrency/pull/104 has shown meaningful speedups for non-allocating futures by removing the rng - perhaps things would be as fast as they are now if you just went back to a `for 0..LEN {}` loop?

matheus-consoli:
I finally got some positive results!

I implemented and tested many variations today, and some were 4.5x slower than `main`.

The biggest improvements came from stopping using `WakerArray`, and not using the `Indexer` type (introduced on #104)

Here're the results:

```
>> critcmp main maybegone-inlinetuple21 -f tuple::join
group             main                                   maybegone-inlinetuple21
-----             ----                                   -----------------------
tuple::join 10    1.06    246.7±0.58ns        ? ?/sec    1.00    233.0±0.66ns        ? ?/sec
```

yoshuawuyts:
@matheus-consoli We probably should keep using the `WakerArray`, even if it costs us some extra performance on the current benchmarks. The reason for that is that our benchmarks only really measure futures with _fast_ `poll` implementations.

At work we've seen cases where calls to `poll` ended up being a bottleneck, which isn't recommended but can absolutely happen if you don't pay super close attention. And in those cases having "perfect" waking will guard against the worst performance implications because it changes the waking algorithm from `O(n^2)` to `O(n)`.

matheus-consoli:
Updated to use `WakerArray`

```
>> critcmp main maybegone-inlinetuple-waking2 -f tuple::join
group             main                                   maybegone-inlinetuple-waking2
-----             ----                                   -----------------------------
tuple::join 10    1.00    246.7±0.58ns        ? ?/sec    2.12    524.0±1.11ns        ? ?/sec
```

yoshuawuyts (approved):

Issue #85, opened 2022-11-15
Deterministically pick the next future to poll


eholk:
This issue is partly about implementing it, but more importantly to have the discussion about whether we want this. There are tradeoffs that I'll try to explain in this issue.

# Fairness and Determinism

This crate strives to be fair in polling futures when there is a choice between which future or stream to poll next. The exact details of the guarantee that we want aren't specified yet (#45), but roughly we want it so that no future is unfairly more likely to complete than another. So if we always polled 0..n, that would be unfair because it would be biased towards futures that are earlier in the list.

The way we guarantee fairness currently is to pick a random index and start polling from there. This keeps any particular entry in the list from getting an unfair advantage. This version does not give us determinism though.

From a [discussion](https://toot.conrad.cafe/@conrad/109344070952043213) with @conradludgate, it seems like it might be enough to just remember the last index we polled and then go from there. This version would give us a deterministic scheduler, but in the long run we don't give an advantage to any index (modulo some special cases I'll get into later).

# Why we want determinism?

Concurrency bugs can be extremely difficult to debug and create test cases for. While there will likely always be nondeterminism coming from external sources (e.g. network requests can take a variable amount of time to complete), we can make it so the scheduler does not introduce any additional nondeterminism. This also makes it possible to carefully construct futures for test cases that can exhibit a particular tricky interleaving that may expose bugs. For debugging, when trying to figure out why something went wrong it can be helpful to replay the exact same sequence of events.

Determinism is also potentially faster, since we can save a few instructions from calculating a random number each time around. I doubt this will matter much in practice though.

# Why do we not want determinism?

As kind of a dual to the testing/debugging argument I made in the last section, nondeterminism can also help with development and testing. It might be that there are weird bugs that only happen in weird interleavings of futures that would be unlikely to come up during development but could happen in production. In that case, introducing nondeterminism can help uncover these. It's basically fuzz testing for the scheduler.

Also, determinism is somewhat in conflict with fairness. If we did something like `[ready(1), ready(2), ready(3)].race()` then by our intuitive definition of fairness we should expect to get either 1, 2, or 3 out. But any deterministic scheduler will always return the same answer. So a deterministic scheduler can only be fair in the long run, but not for short-lived processes.

# What are some other options?

If we decide to have a deterministic scheduler, it might be worth adding a fuzz testing mode that randomizes the scheduling order, which would give us one of the main benefits of a nondeterministic scheduler.

[...]

PR #104, opened 2022-11-21
replace rng with deterministic indexer

yoshuawuyts:
Closes https://github.com/yoshuawuyts/futures-concurrency/issues/85. Additionally it also significantly simplifies our iteration logic by baking it into the `Indexer` struct directly:

```rust
// before
let len = this.streams.len();
let r = this.rng.generate(len as u32) as usize;
for index in (0..len).map(|n| (r + n).wrapping_rem(len)) { ... }

// after
for index in this.indexer.iter() { ... }
```

## Performance

The numbers overall seem pretty comparable, but they spiked a bit on non-allocating APIs which are otherwise fairly simple. `impl Race for array` showed one of the bigger changes:
```text
❯ critcmp main patch
group             main                    patch
-----             ----                    -----
array::race 10    1.22   652.2±24.21ns    1.00   535.9±31.39ns
array::race 100   1.52      4.8±0.27µs    1.00      3.2±0.02µs
array::race 1000  1.07     28.4±0.56µs    1.00     26.6±0.58µs
```


matheus-consoli (changes requested):
I like how this simplifies the iteration, and the strategy seems to make sense to me!

Once we fix the offset increment, we should be good to go

matheus-consoli (approved):
This looks great!

src/utils/indexer.rs
```diff
@@ -0,0 +1,42 @@
...
+
+impl Indexer {
+    pub(crate) fn new(max: usize) -> Self {
+        Self { offset: 0, max }
+    }
+
+    /// Generate a range between `0..max`, incrementing the starting point
+    /// for the next iteration.
+    pub(crate) fn iter(&mut self) -> IndexIter {
+        // Increment the starting point for next time.
```
eholk, line 18: Does it make sense to try to save where we ended up last time and pick up from there?

PR #109, opened 2022-11-29
Initial implementation of `RaceOk` for tuples

matheus-consoli:
Initial impl of `RaceOk` for tuples

something to look at:

I'm using an array as the internal representation of `AggregateError`, which can be surprising as we tend to expect a tuple-ish.

I opted to use the array because 1) arrays are *much* easier to work on compared to macro-based-variadic-tuples; 2) as of this implementation, we're restricting every `Future` to have the same `Output`, ergo every error has the same type.

### DIFF
--- a/src/future/race_ok/mod.rs
+++ b/src/future/race_ok/mod.rs
@@ -1,6 +1,7 @@
 use core::future::Future;
 
 pub(crate) mod array;
+pub(crate) mod tuple;
 pub(crate) mod vec;
 
 /// Wait for the first successful future to complete.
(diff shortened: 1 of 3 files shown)


eholk (approved):

eholk:
Looks good, thanks!

I thing the `AggregateError` thing is okay for now, although I think we should consider changing it. Partly just for symmetry (it's a little weird to get an array of errors from a tuple), but also to leave open the option of allowing different types in the future. That said, allowing different types will still probably be a breaking change since `(a, b).race_ok()` would have to have a return type like `impl Future<Output = Result<Either<A, B>, AggregateError2<A,  B>>>`, and I'm not sure we actually gain much from that much expressibility. At any rate, this seems like a good place to start experimenting.

PR #113, opened 2022-12-22
Fix a possible memory leak on the impl of `Join` for tuples

matheus-consoli:
If we don't poll `JoinX` to completion, we may leak data.


yoshuawuyts (approved):
Thank you so much for tracking this down!

PR #118, opened 2023-01-09
Use a single Arc shared between Wakers in WakerArray and WakerVec

wishawa:
after = this PR
before = #117 
```$ critcmp  after before 
group                after                                  before
-----                -----                                  ------
array::join 10       1.00      2.3±0.12µs        ? ?/sec    1.13      2.6±0.20µs        ? ?/sec
array::join 100      1.00     23.5±0.71µs        ? ?/sec    1.17     27.5±0.39µs        ? ?/sec
array::join 1000     1.10  1521.6±41.26µs        ? ?/sec    1.00  1377.7±21.95µs        ? ?/sec
array::merge 10      1.00      2.3±0.04µs        ? ?/sec    1.39      3.3±0.07µs        ? ?/sec
array::merge 100     1.00     52.5±6.18µs        ? ?/sec    1.09     57.5±2.20µs        ? ?/sec
array::merge 1000    1.00      2.5±0.06ms        ? ?/sec    1.04      2.6±0.07ms        ? ?/sec
array::race 10       1.00  1158.9±19.21ns        ? ?/sec    1.00  1159.8±25.91ns        ? ?/sec
array::race 100      1.00     12.2±0.33µs        ? ?/sec    1.02     12.5±0.48µs        ? ?/sec
array::race 1000     1.02    144.0±5.25µs        ? ?/sec    1.00    140.7±1.84µs        ? ?/sec
tuple::join 10       1.00      2.1±0.04µs        ? ?/sec    1.17      2.5±0.06µs        ? ?/sec
tuple::merge 10      1.00      2.8±0.06µs        ? ?/sec    1.10      3.1±0.07µs        ? ?/sec
tuple::race 10       1.05  1147.9±36.64ns        ? ?/sec    1.00  1089.3±19.36ns        ? ?/sec
vec::join 10         1.00      2.5±0.03µs        ? ?/sec    1.16      2.9±0.09µs        ? ?/sec
vec::join 100        1.00     34.5±1.06µs        ? ?/sec    1.15     39.6±0.80µs        ? ?/sec
vec::join 1000       1.07  1989.7±40.57µs        ? ?/sec    1.00  1863.9±26.66µs        ? ?/sec
vec::merge 10        1.00      2.9±0.13µs        ? ?/sec    1.25      3.6±0.06µs        ? ?/sec
vec::merge 100       1.00     57.3±1.04µs        ? ?/sec    1.28     73.3±1.38µs        ? ?/sec
vec::merge 1000      1.00      3.6±0.07ms        ? ?/sec    1.17      4.2±0.08ms        ? ?/sec
vec::race 10         1.01  1157.2±18.82ns        ? ?/sec    1.00  1147.6±14.78ns        ? ?/sec
vec::race 100        1.03     12.0±0.19µs        ? ?/sec    1.00     11.6±0.23µs        ? ?/sec
vec::race 1000       1.02    141.4±5.12µs        ? ?/sec    1.00    138.5±3.85µs        ? ?/sec
```


src/utils/wakers/shared_arc.rs
```diff
@@ -0,0 +1,92 @@
...
+// We can save space by making size_of::<usize>() wakers share the same slot.
+// With such change, in 64-bit system, the wake_data array/vec would only need ⌈N/8⌉ slots instead of N.
+
+pub(super) trait WakeDataContainer {
+    /// Get the reference of the wake_data slice. This is used to compute the index.
+    fn get_wake_data_slice(&self) -> &[*const Self];
+    /// Called when the subfuture at the specified index should be polled.
+    fn wake_index(&self, index: usize);
+}
+pub(super) unsafe fn waker_for_wake_data_slot<A: WakeDataContainer>(
```
yoshuawuyts, line 41: Can you add a doc comment explaining what this function is intended to do? Also, since it's marked `unsafe`, can you add an `# unsafe` section explaining which invariants must be upheld?

src/utils/wakers/shared_arc.rs
```diff
@@ -0,0 +1,92 @@
+use core::task::{RawWaker, RawWakerVTable, Waker};
+use std::sync::Arc;
+
+// In the diagram below, `A` is the upper block.
```
yoshuawuyts, line 4: I think this explanation can probably use some work. We should assume this is the first thing that people will read about this structure (it was for me). `A` has not yet been referenced, so it's unclear what that refers to. The blocks themselves in the diagram aren't labeled either.

I think a useful exercise might be to try and explain how this works using only words. And then try and explain it using just a diagram. Neither of which should assume that you have access to the code. Right now the words try and explain what the diagram does, and there seems to be an assumption that people have already looked at the code. All of which is a little bit hard to follow.

yoshuawuyts: Maybe making this a `//!`-style comment would be helpful too. That way we can separate the workings of this module from the docs of the individual exports.

src/utils/wakers/shared_arc.rs
```diff
@@ -0,0 +1,92 @@
...
+        let raw = *pointer;
+        Arc::from_raw(raw)
+    }
+    unsafe fn wake<A: WakeDataContainer, const BY_REF: bool>(pointer: *const ()) {
+        let pointer = pointer as *const *const A;
+        let arc = to_arc::<A>(pointer);
+        // Calculate the index
+        let index = ((pointer as usize) // This is the slot our pointer points to.
+            - (arc.get_wake_data_slice() as *const [*const A] as *const () as usize)) // This is the starting address of wake_data.
+            / std::mem::size_of::<*const A>();
```
yoshuawuyts, line 65: Rather than adding comments, can I ask you to assign these to descriptive variables instead? e.g. 
```suggestion
        let pointer_slot = (pointer as usize);
        let wake_data_slice = arc.get_wake_data_slice();
        let start_addr = (wake_data_slice as *const [*const A] as *const () as usize);
        let index = (pointer_slot - start_addr) / std::mem::size_of::<*const A>();
```

I'm not super sure why `std::mem::size_of::<*const A>()` exists, but maybe that too can be assigned to a variable?

yoshuawuyts: Also, now that I've written this out like this - we should probably not perform pointer-to-usize casts. The stdlib has been rewritten to move away from that, meaning we can't PR this code the stdlib right now. Instead we should probably use the [strict provenance polyfill](https://github.com/Gankra/sptr) to perform pointer arithmatic.

src/utils/wakers/vec/waker_vec.rs
```diff
@@ -1,31 +1,74 @@
...
+/// A collection of wakers sharing the same allocation.
 pub(crate) struct WakerVec {
     wakers: Vec<Waker>,
-    readiness: Arc<Mutex<ReadinessVec>>,
+    inner: Arc<WakerVecInner>,
+}
+
+/// See [super::super::shared_arc] for how this works.
+struct WakerVecInner {
+    wake_data: Vec<*const Self>,
```
yoshuawuyts, line 17: I guess I'm somewhat confused why this takes a `Vec<*const Self>`? Previously we had a vec of wakers, and a vec of "readiness". Things are definitely more complex because of this PR, and I think it would be helpful if the structs and fields all spelled out more closely what their intent is.

Because we want to PR these APIs to the stdlib, I want to make sure everything we do in here is documented thoroughly. It isn't just be me who needs to understand it, it needs to be everyone working on the stdlib - and then maintained into perpetuity.

src/utils/wakers/shared_arc.rs
```diff
@@ -0,0 +1,92 @@
...
+
+        arc.wake_index(index);
+
+        // Dropping the Arc would decrement the strong count.
+        // We only want to do that when we're not waking by ref.
+        if BY_REF {
+            std::mem::forget(arc);
```
yoshuawuyts, line 72: Can you elaborate why using `std::mem::forget` doesn't cause a memory leak here?

src/utils/wakers/shared_arc.rs
```diff
@@ -0,0 +1,92 @@
+use core::task::{RawWaker, RawWakerVTable, Waker};
```
yoshuawuyts, line 1: I commented on this before, but in this file we're doing quite a bit of unsafe. I understand we have to use `unsafe fn` to pass it to `RawWakerVtable`. But can we wrap calls to `unsafe` functions in `unsafe {}` blocks, and document why they're actually safe to perform?

Basically I'm asking to follow [RFC 2585](https://github.com/rust-lang/rfcs/blob/master/text/2585-unsafe-block-in-unsafe-fn.md) semantics in this code and Clippy's [safety-doc lint](https://rust-lang.github.io/rust-clippy/stable/index.html#missing_safety_doc), since that will make it easier to maintain and onboard new people with.

src/utils/wakers/shared_arc.rs
```diff
@@ -0,0 +1,92 @@
...
+            std::mem::forget(arc);
+        } else {
+            std::mem::drop(arc);
+        }
+    }
+    unsafe fn drop_waker<A: WakeDataContainer>(pointer: *const ()) {
+        let pointer = pointer as *const *const A;
+        let arc = to_arc::<A>(pointer);
+        // Decrement the strong count by dropping the Arc.
+        std::mem::drop(arc);
```
yoshuawuyts, line 81: Can you use [`Arc::decrement_strong_count`](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.decrement_strong_count) here instead?

src/utils/wakers/shared_arc.rs
```diff
@@ -0,0 +1,92 @@
...
+    }
+    fn create_vtable<A: WakeDataContainer>() -> &'static RawWakerVTable {
+        &RawWakerVTable::new(
+            clone_waker::<A>,
+            wake::<A, false>,
+            wake::<A, true>,
```
yoshuawuyts, line 87: Can these be two separate functions? Given you have a comment within the `wake` function explaining the difference in behavior already, it would be easier to follow if they were to separate functions instead.

yoshuawuyts (changes requested):
Thanks so much for this PR! - I'm trying to better understand everything that's going on, so this first round of reviews is mostly focused on documentation and clarifying what's going on in the unsafe blocks. I've mentioned this in the review comments as well: but the intent of this crate is to upstream parts of it to the stdlib, which means we need to build it with that in mind. I hope that makes sense; thanks!

src/utils/wakers/shared_arc.rs
wishawa: I added `#[deny(unsafe_op_in_unsafe_fn)]` to the function. This lint could be expanded to apply crate-wide in the future.

wishawa:
@yoshuawuyts thanks for the review! I added more documentation/comments, renamed things so that they (hopefully) make more sense, and simplified the code quite a bit. Please take a look!

src/utils/wakers/vec/waker_vec.rs
wishawa: `wake_data` was poor naming on my part. In the latest commit I gave this mechanism the name "redirect" and added explanations in the module doc of `shared_arc`.

src/utils/wakers/shared_arc.rs
wishawa: Added the polyfill. Luckily we weren't doing any usize->pointer cast so there's not much to worry about regarding provenance.

wishawa:
@yoshuawuyts @matheus-consoli is there still interest in this change? I think reducing the number of allocations from N to 1 is very beneficial.

PR #121, opened 2023-02-09
nit: use constant to call recursive macro once instead of several times

poliorcetics:
The `utils::tuple_len!()` macro is now called once instead of 6 times.

Measuring locally with `cargo build --timings --release`, it compiles 0.15 seconds less (from 4.5s to 4.35s), which is not very useful, but it bothered me to call a recursive macro 6 times for no real reason 😅


yoshuawuyts (approved):
hehe, yeah fair enough. Thank you!

Issue #62, opened 2022-11-13
Add `stream::Chain` trait


yoshuawuyts:
This is a core iteration "order of execution" operation, enabling people to author:

**before**
```rust
// Manual sequential iteration.
for n in 0..10 { dbg!(n); }
for n in 11..20 { dbg!(n); }
```
**after**
```rust
// The same semantics using `chain`.
for n in ((0..10), (11..20)).chain() {
    dbg!(n)
}
```
This is particularly useful to do things like "send one last message after EOF".

PR #124, opened 2023-02-09
feat: Implement `Chain` for tuples up to 12 elements

poliorcetics:
I think the index + const pattern I used could be reused for other trait implementations, though I
have not checked.

Also, we could do away with `done`, it's enough to check `this.index == $mod_name::LEN` in the `assert!`, but
without benchmarking I'm not sure which is better, especially since adding a single `bool` is unlikely to be costly
since the futures will probably be much heavier than it.

Helps with #62.

### DIFF
--- a/src/stream/chain/tuple.rs
+++ b/src/stream/chain/tuple.rs
@@ -1,1 +1,133 @@
+use core::fmt;
+use core::pin::Pin;
+use core::task::{Context, Poll};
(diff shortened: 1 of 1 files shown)


poliorcetics:
Trivia: this MR is exactly twice the number of the issue it is working on, which is useless but makes me smile :D

Issue #63, opened 2022-11-13
Add `stream::Zip` trait


yoshuawuyts:
This is a core iteration "order of execution" operation, enabling people to author:

**after**
```rust
let s1 = async_iter::repeat(0).take(5);
let s2 = async_iter::repeat(1).take(5);
let s3 = async_iter::repeat(2).take(5);

for await (n1, n2, n3) in (s1, s2, s3).zip() {
    assert_eq!((n1, n2, n3), (0, 1, 2));
}
```
This is better than the nested tuple approach of non-trait based `zip`.

## Tasks
- [ ] `Zip` trait
- [ ] `impl Zip for vec`
- [ ] `impl Zip for array`
- [ ] `impl Zip for tuple`

yoshuawuyts:
For the `Zip` variant: we may want to skip directly ahead to #22 in terms of implementation, creating an output slot of `Container[MaybUninit<T>]` where we write into, and then `mem::swap` on each iteration.

Metadata would need to be stored in an accompanying `PollState` slot. The main things to keep an eye will be to:

- ensure to implement the `PinnedDrop` trait to drop memory in the case of a cancellation
- switch between `PollState::Pending` and `PollState::Ready` while data is actively being yielded, and only switch to `PollState::Consume` at the end.

The semantics of `Zip` are that if at any point one of the streams yields `Poll::Ready(None)`, all other streams are cancelled. We should create a case where one stream yields more times than the other streams so that we can exercise `miri`'s checkers during CI.

yoshuawuyts:
Filed a concern in https://github.com/rust-lang/rust/issues/80094 since its existence requires us to author:

```rust
let mut s = Zip::zip([a, b, c]);
```

instead of:

```rust
let mut s = [a, b, c].zip();
```

PR #125, opened 2023-02-10
feat: Implement `Zip` for tuples up to 12 elements

poliorcetics:
I think we can do better but I would like my other PRs to be merged before that, to be able
to do the improvements over all of them at once.

Helps with #63


yoshuawuyts (approved):
Amazing; thanks so much!

PR #134, opened 2023-05-09
Make tuple types public

zjijz:
This updates all tuple implementation types of `future::{Join, TryJoin, Race, RaceOk}` and `stream::{Chain, Merge, Zip}` to be publicly exported out of `crate::tuple`, just like all `Vec` and `array` implementations in `crate::vec` and `crate::array`.

Besides for symmetry, my motivation with this change was to allow easier naming of types that use these combinators (without needing to resort to `impl Trait` syntax) and to allow implementing a trait on the concrete implementations for added functionality.

For example, I wanted to write some type like:
```
pub type MathOperator<Fut1, Fut2, F> = Map<Join2<Fut1, Fut2>, F>;
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -82,6 +82,36 @@ pub mod prelude {
 pub mod future;
 pub mod stream;
 
+/// Helper functions and types for tuples.
+pub mod tuple {
+    pub use crate::future::join::tuple::{
+        Join0, Join1, Join10, Join11, Join12, Join2, Join3, Join4, Join5, Join6, Join7, Join8,
+        Join9,
+    };
+    pub use crate::future::race::tuple::{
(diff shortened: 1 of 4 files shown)


yoshuawuyts:
Hi! Thanks for opening this PR. We intentionally don't make the tuple types public because in the stdlib we wouldn't want to either. We're waiting on variadic tuples to land first so we don't have to expose hundreds of new types.

zjijz:
Is there any possibility we could add them to this library and document them to only be a stand-in until variadic generics are available? My feeling is that those are probably still (at least) a few years out so this would make my use case impossible for those years. Could be gated by an `unstable` feature or just removed in a major version revision later.

If `futures-concurrency` gets merged into `std` before that, they could be left out or put under an unstable feature.

PR #136, opened 2023-05-16
Fix #135: check if Future has completed before polling it

### LINKED ISSUE #135: TryJoin2 polls completed futures
Repro:

```rust
use futures_concurrency::future::TryJoin;
use std::future::poll_fn;
use std::task::Poll;

async fn yield_once() -> Result<(), ()> {
    let mut yielded = false;
    poll_fn(|cx| {
        if yielded {
            Poll::Ready(())
        } else {
            yielded = true;
            cx.waker().wake_by_ref();
            Poll::Pending
        }
    })
    .await;
    Ok(())
}

fn main() {
    futures::executor::block_on((async { Ok(()) }, async { yield_once().await }).try_join())
        .unwrap();
}
```

Output:

```
thread 'main' panicked at '`async fn` resumed after completion', src/main.rs:21:34
```

matheus-consoli:
Fix #135, we're missing the check before polling.


yoshuawuyts (approved):
Fantastic, thank you!

Issue #137, opened 2023-06-11
`tracing` can't track the execution duration of futures passed to `futures-concurrency`


yoshuawuyts:
An example provided by @jstarks:

```rust
let fut1 = async { ... }.instrument(info_span!("span1"));
let fut2 = async { ... }.instrument(info_span!("span2"));
(fut1, fut2).join();
```

The problem is that both spans are always measured to have the same duration, because the wrapping future returned by `.instrument` keeps the span alive until the future is dropped, not until the future's `poll()` returns ready.

On our end the change we can make is to more eagerly drop futures on completion, rather than only dropping them when the final future resolves. Conservatively tagging this as a bug, since the resulting behavior is not right - even if we aren't failing to uphold any of Rust's async invariants.

PR #138, opened 2023-06-11
Drop futures as soon as they're done for `array::join`

yoshuawuyts:
Makes progress towards #137. If we accept this fix we'll want to copy it for all our APIs.

### DIFF
--- a/src/future/join/array.rs
+++ b/src/future/join/array.rs
@@ -1,5 +1,5 @@
 use super::Join as JoinTrait;
-use crate::utils::{self, array_to_manually_drop, PollArray, WakerArray};
+use crate::utils::{self, FutureArray, PollArray, WakerArray};
 
 use core::array;
 use core::fmt;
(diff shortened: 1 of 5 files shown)


src/future/join/array.rs
```diff
@@ -29,7 +29,7 @@ where
     wakers: WakerArray<N>,
     state: PollArray<N>,
     #[pin]
-    futures: [Fut; N],
+    futures: [MaybeUninit<Fut>; N],
```
jstarks, line 32: Would `ManuallyDrop` be clearer?

src/future/join/array.rs
```diff
@@ -94,20 +94,32 @@ where
...
+                        .map_unchecked_mut(|t| t.assume_init_mut())
+                        .poll(&mut cx)
+                } {
                     this.items[i] = MaybeUninit::new(value);
                     this.state[i].set_ready();
                     *this.pending -= 1;
                 }
 
+                // If the state was changed from "pending" to "ready", drop the future.
+                if this.state[i].is_ready() {
```
jstarks, line 118: Seems like this could merge into the previous `if` block.

src/future/join/array.rs
yoshuawuyts: Yes, that's a great suggestion — it would be!

src/future/join/array.rs
yoshuawuyts: I initially had that, but it gave a lifetime error. Let me change it to use ManuallyDrop and try again.

yoshuawuyts:
Updated with feedback from review! I'm pretty happy with the logic as it is, and I think we could merge it. But I think I'd be happiest if we could factor out the manual `[ManuallyDrop<T>; N]` logic into its own file, because we'll want to share it between all array implementations.

PR #142, opened 2023-06-12
Drop futures as soon as they're done for `{array, vec}::try_join`

yoshuawuyts:
Makes progress towards #137. Thanks!


yoshuawuyts:
The MacOS runners seem to be timing out immediately, which probably means GitHub is experiencing issues somewhere. We're not doing anything MacOS specific though, so we should be able to safely ignore that.

PR #143, opened 2023-06-21
tuple::join - drop futures in-place

yoshuawuyts:
Reference: #137. This implements drop-in-place semantics for `tuple::join`. Thanks!


yoshuawuyts:
oh apparently the nightly linux toolchain is broken so miri fails to install. I've run miri locally and it seems to pass - given Miri is a virtualized environment, that should be good enough, so allowing this to be merged!

PR #145, opened 2023-06-23
`tuple::try_join` - drop futures in-place

yoshuawuyts:
Reference: #137. This is the last of the batch, implementing "drop-in-place" semantics for `tuple::try_join`. Thanks!


yoshuawuyts:
I'm feeling good about this - we added a miri test to this as well and it's passing. Going to hit merge on this and publish a new release.

PR #147, opened 2023-07-17
Add a new `future::FutureGroup` type for dynamic async concurrency

yoshuawuyts:
This is similar to `FuturesUnordered` from the futures library, or `JoinSet` from tokio.

# Example

```rust
let mut set = FutureGroup::new();
group.insert(future::ready(2));
group.insert(future::ready(4));

let mut out = 0;
while let Some(num) = group.next().await {
    out += num;
}
assert_eq!(out, 6);
assert_eq!(group.len(), 0);
assert!(set.is_empty());
```

## Tasks
- [x] Initial implementation
- [ ] Make use of an internal `WakerSet` to improve performance
- [x] Figure out a solution to our lending iterator problem

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,7 @@ bitvec = { version = "1.0.1", default-features = false, features = ["alloc"] }
 futures-core = "0.3"
 pin-project = "1.0.8"
 slab = "0.4.8"
+async-iterator = "2.1.0"
 
 [dev-dependencies]
 async-channel = "1.9.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ pin-project = "1.0.8"
 slab = "0.4.8"
 
 [dev-dependencies]
+async-channel = "1.9.0"
 futures = "0.3.25"
 futures-lite = "1.12.0"
 criterion = { version = "0.3", features = ["async", "async_futures", "html_reports"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,6 @@ bitvec = { version = "1.0.1", default-features = false, features = ["alloc"] }
 futures-core = "0.3"
 pin-project = "1.0.8"
 slab = "0.4.8"
-async-iterator = "2.1.0"
 
 [dev-dependencies]
 async-channel = "1.9.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ harness = false
 bitvec = { version = "1.0.1", default-features = false, features = ["alloc"] }
 futures-core = "0.3"
 pin-project = "1.0.8"
+slab = "0.4.8"
 
 [dev-dependencies]
 futures = "0.3.25"
--- a/src/future/future_set.rs
+++ b/src/future/future_set.rs
@@ -11,7 +11,7 @@ use std::task::Poll;
 
 /// A handle to a dynamic set of futures.
 ///
-/// See [`FutureSet`] for more.
+/// This is returned by calling the [`handle`][`FutureSet::handle`] method on [`FutureSet`].
 #[derive(Clone)]
 #[pin_project]
 pub struct FutureSetHandle<T> {
--- a/src/future/future_set.rs
+++ b/src/future/future_set.rs
@@ -1,8 +1,7 @@
-use futures_core::Stream;
-use pin_project::pin_project;
+use async_iterator::LendingIterator;
 use slab::Slab;
 use std::fmt::{self, Debug};
-use std::future::Future;
+use std::future::{self, Future};
 use std::pin::Pin;
 use std::task::Poll;
 
@@ -12,6 +11,7 @@ use std::task::Poll;
 ///
 /// ```rust
 /// use futures_concurrency::future::FutureSet;
+/// use async_iterator::LendingIterator;
 /// use futures_lite::StreamExt;
 ///
 /// # futures_lite::future::block_on(async {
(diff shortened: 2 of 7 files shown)


yoshuawuyts:
Actually, now that I think of it: we probably should remove the `Handle` in favor of using a channel.

yoshuawuyts:
Okay, done - removed the handle! There is still a real question about how to weave through the handle pattern though. I briefly tried this test case, but it doesn't work out because of the lifetimes:

```rust
    #[test]
    fn concurrent_channel() {
        enum Message<T> {
            Future(Pin<Box<dyn Future<Output = T> + 'static>>),
            Output(T),
        }
        futures_lite::future::block_on(async {
            let mut set = FutureSet::new();

            let (sender, receiver) = async_channel::bounded(1);
            sender.try_send(async { 2 + 2 }).unwrap();
            let receiver = receiver.map(|t| Message::Future(Box::pin(t)));

            set.insert(async { Message::Output(1 + 1) });
            let mut stream = (set, receiver).merge();

            let mut out = 0;
            while let Some(msg) = stream.next().await {
                match msg {
                    Message::Future(fut) => set.insert(async move { // `set` has been moved
                        let output = fut.await;
                        Message::Output(output)
                    }),
                    Message::Output(num) => out += num,
                }
            }

            assert_eq!(out, 6);
        });
    }
```

I'm pretty certain there is some formulation of `LendingAsyncIterator` which will allow us to pull through the underlying sources on each iteration so they can be modified from inside the loop. I wrote [an example of this](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=92cb8f410bf782c3c9e59c47c947a0bc) on nightly a few weeks ago, which is [currently broken again](https://github.com/rust-lang/rust/issues/113796). With that we could do something like this:

```rust
let mut iter = vec![0; 1]).into_async_iter().lend_mut();

let mut appended = false;
while let Some((item, vec)) = iter.next().await {
    if !appended {
        vec.push(item + 1);
        appended = true;
    }
    println!("{item}");
}
```

We should explore some options with this:

- Add a `StreamSet` type which has both `iter` and `lending_iter` sets of methods - the latter of which borrows the underlying iterators and returns them. This would be a new trait, which only operates on `Unpin` async iterators.
- Do what we did in [the playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=92cb8f410bf782c3c9e59c47c947a0bc) and add an `AsyncIterator::lend` method which derefs the underlying stream into its components.
    - We should combine this with `impl Deref{,Mut} for TupleMerge where A..: Unpin {}` or whatever. So if you `.merge` N streams, you can `.lend` them to extract the individual underlying streams.

matheus-consoli:
@yoshuawuyts, am I missing something, or does adding a `'static` bound to the generic `AsyncIterator` fix the problem on your playground example?

```rust
- impl<I: AsyncIterator          > AsyncLendingIterator for LendMut<I>
+ impl<I: AsyncIterator + 'static> AsyncLendingIterator for LendMut<I>

- impl<I: AsyncIterator          > AsyncLendingIterator for Lend<I>
+ impl<I: AsyncIterator + 'static> AsyncLendingIterator for Lend<I>
```

yoshuawuyts:
@matheus-consoli we explicitly want the lifetime to be non-static. I filed https://github.com/rust-lang/rust/issues/113796 which has just been fixed, so this should work again on the next nightly.

yoshuawuyts:
Okay, updated it! - This now seems to work as expected, and I'm pretty happy with it. Oh CI is probably failing because this requires nightly features. If we want to move forward with this PR we should apply the proper cfgs for this to work.

## Examples

Here are two examples of this in action:
```rust
//! A basic example showing how to add items to the set

let mut set = FutureSet::new();
set.insert(async { 1 + 1 });
set.insert(async { 2 + 2 });

let mut out = 0;
while let Some((num, _set)) = set.next().await {
    out += num;
}
assert_eq!(out, 6);
assert_eq!(set.len(), 0);
assert!(set.is_empty());
```

```rust
//! A more complex example, showing how to concurrently await both a channel of futures, as well as the futures received on that channel.
let mut set = FutureSet::new();

let (sender, receiver) = async_channel::bounded(1);
sender.try_send(async { 2 + 2 }).unwrap();

set.insert(async { Message::Output(1 + 1) });
set.insert(async move {
    let fut = receiver.recv().await.unwrap();
    Message::Future(Box::pin(fut))
});

let mut out = 0;
while let Some((msg, set)) = set.next().await {
    match msg {
        Message::Future(fut) => set.insert(async move {
            let output = fut.await;
            Message::Output(output)
        }),
        Message::Output(num) => out += num,
    }
}

assert_eq!(out, 6);
```

## Further Improvements
It seems though that directly implementing `LendingIterator` for `FutureSet` is a bit clumsy, and instead we'd indeed be better off implementing the generalized `.lend` method. Like, we're basically hard-coding that now, which we means we're taking an ergonomics penalty every time someone doesn't want to update from the loop. With that we could rewrite the core loop of the first example to not require the inert `_set` item:

```rust
let mut set = FutureSet::new();
set.insert(async { 1 + 1 });
set.insert(async { 2 + 2 });

let mut out = 0;
while let Some(num) = set.next().await {
    out += num;
}
```

yoshuawuyts:
`async-iterator@2.2.0` now has support for [the `lend` and `lend_mut` methods](
https://github.com/yoshuawuyts/async-iterator/blob/30497767b1d29c116484bd18c39beaf3d0bc0032/src/iter/mod.rs#L46-L64) which will allow us to simplify some of this.

yoshuawuyts:
Updated the implementation to take homogenous future types only. This saves intermediate allocations, and allows users to choose their own strategy for unifying types.

It's not without tradeoffs, but those tradeoffs seem inherent to Rust's handling of heterogenous types, and are shared by e.g. unifying the output types of `Stream::merge` as well.

yoshuawuyts:
We should reimplement this on top of https://github.com/yoshuawuyts/futures-concurrency/pull/149. Either by copying the impl (which should be simple enough), or by using it internally and wrapping types in `stream::once`. A lot of the hard work has gone into that PR, and we just have to adapt it for our uses.

yoshuawuyts:
Closing in favor of #150

PR #149, opened 2023-08-09
Add the `StreamSet` type

yoshuawuyts:


yoshuawuyts:
Spent the day debugging performance, and I think we're done now. This impl is now on par with the futures-rs version, but with way fewer unsafe low-level concurrency constructs. All that's left is providing a `.lend` method which is compatible with the `Stream` trait - probably through a third-party crate.

## Benchmarks

Screenshot 2023-08-12 at 01 18 18

Issue #68, opened 2022-11-14
Add tests for `no_std`


yoshuawuyts:
The traits themselves are compatible with `no_std`, and the implementations for `array` and `tuple` should be compatible with `no_std` as well. But as we're adding optimizations, we're using things like `bitvec` to store metadata in, speeding things up - but that may require allocations.

We should author tests to ensure that the traits keep working on `no_std` environments, even as we add optimizations. On the implementation side we may even want to swap the implementations we provide for `no_std` environments if it turns out our more optimized approaches are incompatible - but I suspect we may not need to.

Either way: the first step is to author tests and put them in CI so we can ensure that `no_std` works going forward.

yoshuawuyts:
Yes please!

alexmoon:
FYI, I don't think #159 actually does anything. I believe the `#![no-std]` directive in the test file just prevents use of the standard library in the test, but the lib still pulls in std since it does not have a `#![no-std]` directive. I've opened #163 to add no-std support to the crate. On that branch you can run `cargo test --no-default-features` to run no-std tests.

yoshuawuyts:
I believe #163 closes this. Thanks!

PR #159, opened 2023-10-07
Implement no_std tests

jmintb:
Hi, this is a draft PR for #68 to see if I am going in the right direction.

I have taken a test for the `Join` implementation on `array` and copied it into a no_std environment. I believe this should validate that `Join` works in a `no_std` environment. 

Am I correct in assuming that a single test for `Join` is sufficient? as we are validating that traits work in no_std not specific implementations.

Let me know if this is the sort of thing you had in mind :)

### DIFF
--- a/tests/no_std.rs
+++ b/tests/no_std.rs
@@ -1,0 +1,96 @@
+#![no_std]
+
+use core::future;
+use futures_concurrency::{array::AggregateError, prelude::*};
+use futures_lite::future::block_on;
+use futures_lite::prelude::*;
+use futures_lite::stream;
+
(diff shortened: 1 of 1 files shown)


yoshuawuyts:
This seems great; yes please keep going in this direction!

jmintb:
Alright, there is now one test for each trait. I again just reused existing tests so I could easily have missed something. let me know what you think :)

yoshuawuyts (approved):
Oh sweet, thank you! - I was thinking we should add these to CI somehow, but they already do. This looks great; thank you!

PR #161, opened 2023-11-06
use Waker::clone_from internally

### LINKED ISSUE #160: Use `Waker::clone_from` internally
Now that https://github.com/rust-lang/rust/pull/96979 has been merged, we should use `Waker::clone_from` in our `set_waker` functions to remove reduce unnecessary clones. If done right, then from now on we'll typically only clone-on-wake the first time our concurrency types are polled.

This should be relatively simple to implement, so anyone who wants to pick it up should feel free to. It would also be interesting to bench this on nightly, which should have these changes starting ~tomorrow.

miguelraz:
Fixes #160 

Thanks for putting up a nice beginner issue 👍🏾

### DIFF
--- a/src/utils/channel.rs
+++ b/src/utils/channel.rs
@@ -30,7 +30,10 @@ impl<T> Stream for LocalReceiver<T> {
                 if channel.closed {
                     Poll::Ready(None)
                 } else {
-                    channel.waker = Some(cx.waker().clone());
+                    match &mut channel.waker {
+                        Some(prev) => prev.clone_from(cx.waker()),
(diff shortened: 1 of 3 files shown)


yoshuawuyts (approved):
Thank you!

yoshuawuyts:
Screenshot 2023-11-07 at 14 57 14

It seems this has improved performance roughly 3-30% on all tests. The median improvement looked like it was around 5%. But in some cases it produced fun graphs like the one above. Thank you for this!

miguelraz:
That's rad!

PR #163, opened 2024-01-28
Add no-std support

alexmoon:
Adds `std` and `alloc` features to enable support for no-std targets.

Mostly this was just a matter of updating imports from `std` to `core` or `alloc` as necessary along with `cfg` directives to omit things like the vec implementations that require an allocator.

The only real issue I ran into was that the implementations of `WakerArray` and `WakerVec` required both `Arc` and `Mutex`. `Arc` is only available on `alloc` and `Mutex` is only available on `std`. So I've added very basic no-std implementations of both which simply revert to using the parent waker directly instead of trying to do precise waking. I think it should be possible to implement precise waking in a no-std environment, but will require messing with `RawWakers` and probably pinning and a fair amount of unsafe code. This does mean that on no-std targets all futures will be polled on every wake.

The only other behavior difference is removing printing of the child error sources in the `Debug` impl of the `AggregateError` structs. This was necessary because the `Error` trait is currently std only. It does have the side benefit of there now being a `Debug` impl for any `AggregateError` whose `T` impls `Display` instead of only when `T` impls `Error`.

I updated CI to run `cargo check` on std, alloc, and core configurations. Let me know if you want CI to run `cargo test` on the no-std configurations as well.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -34,6 +34,18 @@ jobs:
         command: check
         args: --all --bins --examples
 
+    - name: check no-std
+      uses: actions-rs/cargo@v1
+      with:
+        command: check
+        args: --all --no-default-features
+
+    - name: check alloc
+      uses: actions-rs/cargo@v1
+      with:
+        command: check
+        args: --all --no-default-features --features alloc
+
     - name: tests
       uses: actions-rs/cargo@v1
       with:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,12 +27,17 @@ harness = false
 name = "compare"
 harness = false
 
+[features]
+default = ["std"]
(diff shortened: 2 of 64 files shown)


max-heller:
I'm excited for this as I'd like to use this crate in a no-std, minimal-alloc environment. Even if precise waking isn't possible, it'd be great to have access to these combinators workout having to reinvent the wheel!

yoshuawuyts (approved):
This looks fantastic, thank you!

Apologies for taking a minute to review this; I was out of office all of December and this fell off my review queue. I'm very grateful for this PR though, as support for `no_std` and `alloc` environments is really important! I've gone and filed https://github.com/yoshuawuyts/futures-concurrency/issues/166 as well to make sure we eventually improve the performance as well.

PR #164, opened 2024-02-16
Concurrent stream

yoshuawuyts:
This implements a system for concurrent handling of items in streams.

Pitch: we don't want to ever work with a `Stream<Item = impl Future>` for concurrency; what we really want is a `ConcurrentAsyncIterator` trait which does all the right things for us. That means the solution to "barbara battles buffered streams" is to tell her not to put futures inside of streams, and instead use "concurrent streams" directly.

This is likely a more robust direction than trying to weave a new `poll_bg` / `poll_progress` method through every future and async iterator. If we do things right, resolving this issue will come at the overhead of roughly an extra copy.

# todos

- [x] base function impl based on `StreamGroup`
- [x] make the `for_each` closure return a future (T_T can't use nightly features here)
- [x] write integration tests to validate the impl
- [x] split it into a type with methods on it to set the values
- [x] convert that into own trait
- [x] add a drain method
- [x] add a conversion from a regular stream
- [x] implement a `map` method
- [x] implement a `for_each` method
- [x] split `limit` out into its own method
- [x] `try_for_each` to short-circuit on error
- [x] leverage a private `Try` family of traits
- [x] implement a `collect` method
- [x] stack pin the internal `FutureGroup`

## References

- [Tyler's post](https://tmandry.gitlab.io/blog/posts/for-await-buffered-streams/)
- [Async Rust user story](https://rust-lang.github.io/wg-async/vision/submitted_stories/status_quo/barbara_battles_buffered_streams.html)
- [Alice's idea](https://www.reddit.com/r/rust/comments/18czxf1/comment/kchm9c9/)
- [Boats' post elaborating on that idea](https://without.boats/blog/poll-progress/)


yoshuawuyts:
This should be good for review now I reckon. It's all implemented and working, just not fully documented and maybe,, smooth? But yeah, I think it's good now for people to start taking a look!

yoshuawuyts:
Oh I guess maybe one more TODO item we could do here is remove the `Pin<Box>` from some of the impls. I think that will require pinning the entire thing, but that's probably fine. This is a consequence of futures being pinned, which we're storing inside of a `FutureGroup`. A little annoying, but probably OK.

yoshuawuyts:
Oh and for anyone wondering about CI: nightly is currently broken because of https://github.com/rust-lang/rust/issues/121708.

yoshuawuyts:
@matheus-consoli I think this PR should be ready to be merged - can I ask you to work your magic to enabled CI to pass?

matheus-consoli:
:magic_wand: simsalabim faça o CI funcionar pra mim :magic_wand:

PR #167, opened 2024-03-14
Add `{Future,Stream}::wait_until`

yoshuawuyts:
Makes progress towards lifting some of the items from `futures-time` into this crate: https://github.com/yoshuawuyts/futures-time/issues/5#issuecomment-1656982002

Critically this does not introduce any additional runtime dependencies; it's mostly just new control flow constructs based on the `race` concurrency primitive.


yoshuawuyts:
Both clippy and miri run on nightly, which is currently failing because of https://github.com/rust-lang/rust/issues/121708. We should probably wait for that to resolve before merging.

PR #172, opened 2024-03-22
Better capacity and growth rate for `{Stream,Future}Group`

### LINKED ISSUE #169: Implement better capacity in our group structures
https://github.com/yoshuawuyts/futures-concurrency/pull/168 fixes the immediate insertion time issues of `FutureGroup::insert` - however the code is, how should I say this, not great?

We need to implement our own exponential capacity algorithm. This means implementing a `reserve` method which ensures there is enough capacity. And then making sure that if we run out of capacity we always double our space. We may even want to rename some of our existing `resize` methods to `reserve`, since they don't take items where [`Vec::resize`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.resize) does.

This probably needs to be an immediate follow-up to #168; I don't think we should release a new version without it.

matheus-consoli:
Fixes the same problem as #168 by keeping up with the capacity and growth rate of the group.

When a group tries to grow over its capacity, we double the previous capacity.

`FutureGroup` and `StreamGroup` are structurally very similar, they only differ on the `Stream` implementation, so I isolated the common structure to reduce the duplicated code.

## Performance

Compared to the main branch, we gained 99,775% using [this benchmark](https://github.com/notgull/futures-concurrency-benchmark) (with 1_000_000 futures).

Before:
```rust
group/futures_concurrency::FutureGroup
                        time:   [27.222 s 28.561 s 29.720 s]
```

After:

```rust
group/futures_concurrency::FutureGroup
                        time:   [68.507 ms 68.598 ms 68.752 ms]
                        change: [-99.769% -99.760% -99.748%] (p = 0.00 < 0.05)
                        Performance has improved.
```

## todos
- [x] make `FutureGroup` and `StreamGroup` type aliases for a common struct generic over the poll behavior
- [x] rename `resize` to `reserve` and expose it
- [x] tests and docs and benchmarks?

Closes #169

### DIFF
--- a/src/collections/group_inner.rs
+++ b/src/collections/group_inner.rs
@@ -48,7 +48,7 @@ impl<A> GroupInner<A> {
 
     pub fn insert(&mut self, item: A) -> Key {
         if !self.has_capacity() {
-            self.resize(self.cap + 1 * GROUP_GROWTH_FACTOR);
+            self.resize((self.cap + 1) * GROUP_GROWTH_FACTOR);
         }
 
         let index = self.items.insert(item);
@@ -65,7 +65,7 @@ impl<A> GroupInner<A> {
     pub fn insert_pinned(mut self: Pin<&mut Self>, item: A) -> Key {
         if !self.has_capacity() {
             let r = unsafe { &mut self.as_mut().get_unchecked_mut() };
-            r.resize(r.cap + 1 * GROUP_GROWTH_FACTOR);
+            r.resize((r.cap + 1) * GROUP_GROWTH_FACTOR);
         }
 
         let mut this = self.project();
--- a/src/collections/group_inner.rs
+++ b/src/collections/group_inner.rs
@@ -1,0 +1,144 @@
+use core::{pin::Pin, task::Waker};
+
+use alloc::{collections::BTreeSet, fmt};
+use slab::Slab;
+
+use crate::utils::{PollVec, WakerVec};
+
+const GROUP_GROWTH_FACTOR: usize = 2;
+
+#[pin_project::pin_project]
+pub struct GroupInner<A> {
+    #[pin]
+    pub items: Slab<A>,
+    pub wakers: WakerVec,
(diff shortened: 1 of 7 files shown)


yoshuawuyts:
@matheus-consoli apologies for taking a little while to respond; I was out all of last week. I love that you've implemented a more robust version of #168, and I'm 100% on board with solving #169 the way you've done it. However, I'm unsure whether using the `InnerGroup` implementation you've provided here is the way to do it. Let me try and explain.

In the current implementation we have two structs:
- `StreamGroup` for `T: Stream`
- `FutureGroup` for `T: Future`

In the implementation provided by this PR we have a hierarchy of structs, at the top level there is:
- `StreamGroup` for `T: Stream`
- `FutureGroup` for `T: Future`

Which internally uses a shared impl:

- `InnerGroup<B: PollBehavior>`

`PollBehavior` here abstracts over the differences between `Stream` and `Future`, and is in turn implemented on two additional structs:

- `impl PollBehavior for PollFuture`
- `impl PollBehavior for PollStream`

My worry is that this creates a funnel-shaped abstraction: we're specializing behavior at the top of the abstraction, going through a single shared type, and then specializing again at the bottom of the abstraction. I worry that this may work because `poll_next` is almost identical to `poll`, but if we wanted to add something more left-field such as [`AsyncIteratorGroup`](https://docs.rs/async-iterator/latest/async_iterator/) this layering would break down.

I think for that reason we might be better served with just duplicating the strategy between both impls. And perhaps we can explore ways to share more logic between impls without specializing at the lower levels in future PRs? Would that be ok with you?

matheus-consoli:
sure, i agree with you

my initial idea was to write a single type `InnerGroup` and have `{Future, Stream}Group` be type aliases (that would, more or less, break the funnel-shaped problem), but i also didn't want to expose those in the public API... and i wasn't considering an `AsyncIteratorGroup`.

I will close this PR and open a simpler one.

PR #174, opened 2024-04-06
Update docs for docs.rs

yoshuawuyts:
This updates the docs on docs.rs, giving a better overview of what this library is about and what value it provides. Thanks!

## Screenshot

Screenshot 2024-04-06 at 13-51-12 futures_concurrency - Rust

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,4 +1,6 @@
-//! Performant, portable, structured concurrency operations for async Rust. It works with any runtime, does not erase lifetimes, always handles cancellation, and always returns output to the caller.
+//! Performant, portable, structured concurrency operations for async Rust. It
+//! works with any runtime, does not erase lifetimes, always handles
+//! cancellation, and always returns output to the caller.
 //!
 //! `futures-concurrency` provides concurrency operations for both groups of futures
 //! and streams. Both for bounded and unbounded sets of futures and streams. In both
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -1,62 +1,152 @@
...
+//! | __Concurrent input evaluation__ | `Stream::zip`                    | `Stream::merge`                  |
+//!
+//!  ‡: _This could be addressed by a hypothetical `Stream::unzip` operation,
+//!  however because we aspire for semantic compatibility with `std::iter::Iterator`
+//!  in our operations, the path to adding it is currently unclear_.
+//!  
+//! The following streams implementations are provided by `futures-concurrency`:
+//!
+//! - [`StreamGroup`][stream::StreamGroup]: A growable group of streams which operate as a single unit.
+//! - [`ConcurrentStream`][concurrent_stream::ConcurrentStream]: An asynchronous stream which can concurrently process items.
```
matheus-consoli, line 113: I think the difference between these two may be a bit unclear, especially for those unfamiliar with the lore from The Battle of Barbara. It's also unclear that `ConcurrentStream` is a trait, not a type (it is listed in an "implementations" list). 

We may want to give a short description of the functionality, plus a quick hint on when the user should reach for `.co()`. wdyt?

And (later) we can write a richer doc for `ConcurrentStream` and its mechanisms

PR #177, opened 2024-04-12
Manually track capacity of groups

### LINKED ISSUE #169: Implement better capacity in our group structures

matheus-consoli:
the spiritual successor of #172 (a much simpler implementation)

The problem with using the `Slab` capacity as the group capacity is that the slab has its own growth rate, which differs from the other internal structures. It was possible that the slab had the capacity for more items, but the rest of the structures didn't (I think this was the problem in #168).

The solution for this is explicitly tracking the capacity inside the group.

This also provides a new `reserve` method on `FuturesGroup` and `StreamGroup`, so the user can prepare the group beforehand to receive more items.

Closes #169

### DIFF
--- a/src/future/future_group.rs
+++ b/src/future/future_group.rs
@@ -67,14 +67,15 @@ pub struct FutureGroup<F> {
     wakers: WakerVec,
     states: PollVec,
     keys: BTreeSet<usize>,
+    capacity: usize,
 }
 
 impl<T: Debug> Debug for FutureGroup<T> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.debug_struct("FutureGroup")
             .field("slab", &"[..]")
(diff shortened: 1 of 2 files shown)


yoshuawuyts (approved):
this looks great; thank you!

PR #179, opened 2024-04-16
add new benchmark for `FutureGroup`

soooch:
This benchmark aims to measure latency between futures becoming ready and the stream producing them.
The bench currently shows `FutureGroup` trailing significantly behind `FuturesUnordered` for larger test sizes.

```
~/c/futures-concurrency (new-future-group-bench)> cargo bench future_group_poll_latency
    Finished bench [optimized + debuginfo] target(s) in 0.05s
     Running benches/bench.rs (target/release/deps/bench-26d01e1345f7786c)
Gnuplot not found, using plotters backend
future_group_poll_latency/FutureGroup/Params { init_size: 10, pct_ready_per_round: 0.001 }                                                                             
                        time:   [630.81 ns 631.56 ns 632.24 ns]
                        change: [+7.1846% +7.2671% +7.3558%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 7 outliers among 100 measurements (7.00%)
  7 (7.00%) high mild
future_group_poll_latency/FuturesUnordered/Params { init_size: 10, pct_ready_per_round: 0.001 }                                                                             
                        time:   [466.72 ns 469.05 ns 471.58 ns]
                        change: [-1.7131% -1.2735% -0.8808%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 17 outliers among 100 measurements (17.00%)
  14 (14.00%) low severe
  3 (3.00%) low mild
future_group_poll_latency/FutureGroup/Params { init_size: 10, pct_ready_per_round: 0.2 }                                                                             
                        time:   [537.25 ns 539.50 ns 541.62 ns]
                        change: [+2.5562% +2.9383% +3.2572%] (p = 0.00 < 0.05)
                        Performance has regressed.
future_group_poll_latency/FuturesUnordered/Params { init_size: 10, pct_ready_per_round: 0.2 }                                                                             
                        time:   [366.65 ns 368.35 ns 369.97 ns]
                        change: [-2.7245% -2.4190% -2.0874%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 30 outliers among 100 measurements (30.00%)
  16 (16.00%) low severe
  4 (4.00%) low mild
  8 (8.00%) high mild
  2 (2.00%) high severe
future_group_poll_latency/FutureGroup/Params { init_size: 10, pct_ready_per_round: 1.0 }                                                                             
                        time:   [404.96 ns 410.50 ns 416.19 ns]
                        change: [+13.031% +13.890% +14.650%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 17 outliers among 100 measurements (17.00%)
  16 (16.00%) low severe
  1 (1.00%) low mild
future_group_poll_latency/FuturesUnordered/Params { init_size: 10, pct_ready_per_round: 1.0 }                                                                             
                        time:   [276.28 ns 276.34 ns 276.39 ns]
                        change: [-0.7471% -0.7059% -0.6578%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 13 outliers among 100 measurements (13.00%)
  7 (7.00%) low mild
  4 (4.00%) high mild
  2 (2.00%) high severe
future_group_poll_latency/FutureGroup/Params { init_size: 100, pct_ready_per_round: 0.001 }                                                                             
                        time:   [14.159 µs 14.207 µs 14.255 µs]
                        change: [+4.3452% +4.4973% +4.6408%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 18 outliers among 100 measurements (18.00%)
  8 (8.00%) low severe
  7 (7.00%) low mild
  1 (1.00%) high mild
  2 (2.00%) high severe
future_group_poll_latency/FuturesUnordered/Params { init_size: 100, pct_ready_per_round: 0.001 }                                                                             
                        time:   [4.9547 µs 4.9589 µs 4.9646 µs]
                        change: [-1.6303% -1.2868% -0.9175%] (p = 0.00 < 0.05)
                        Change within noise threshold.
future_group_poll_latency/FutureGroup/Params { init_size: 100, pct_ready_per_round: 0.2 }                                                                             
                        time:   [11.181 µs 11.182 µs 11.184 µs]
                        change: [+2.6277% +2.6520% +2.6760%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 7 outliers among 100 measurements (7.00%)
  1 (1.00%) low mild
  6 (6.00%) high mild
future_group_poll_latency/FuturesUnordered/Params { init_size: 100, pct_ready_per_round: 0.2 }                                                                             
                        time:   [2.9196 µs 2.9202 µs 2.9207 µs]
                        change: [-1.6521% -1.6239% -1.5951%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) low mild
  1 (1.00%) high mild
future_group_poll_latency/FutureGroup/Params { init_size: 100, pct_ready_per_round: 1.0 }                                                                             
                        time:   [4.0768 µs 4.0777 µs 4.0787 µs]
                        change: [-0.3427% +0.0016% +0.2321%] (p = 1.00 > 0.05)
                        No change in performance detected.
Found 5 outliers among 100 measurements (5.00%)
  1 (1.00%) low severe
  1 (1.00%) low mild
  3 (3.00%) high mild
future_group_poll_latency/FuturesUnordered/Params { init_size: 100, pct_ready_per_round: 1.0 }                                                                             
                        time:   [2.7793 µs 2.7796 µs 2.7801 µs]
                        change: [-6.4224% -6.2342% -6.0593%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 9 outliers among 100 measurements (9.00%)
  1 (1.00%) low severe
  8 (8.00%) high severe
Benchmarking future_group_poll_latency/FutureGroup/Params { init_size: 1000, pct_ready_per_round: 0.001 }: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 6.4s, enable flat sampling, or reduce sample count to 60.
future_group_poll_latency/FutureGroup/Params { init_size: 1000, pct_ready_per_round: 0.001 }                                                                             
                        time:   [1.0431 ms 1.0460 ms 1.0488 ms]
                        change: [-0.7482% -0.5783% -0.3925%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 20 outliers among 100 measurements (20.00%)
  1 (1.00%) high mild
  19 (19.00%) high severe
future_group_poll_latency/FuturesUnordered/Params { init_size: 1000, pct_ready_per_round: 0.001 }                                                                            
                        time:   [52.312 µs 52.334 µs 52.362 µs]
                        change: [-0.0070% +0.0275% +0.0639%] (p = 0.14 > 0.05)
                        No change in performance detected.
Found 7 outliers among 100 measurements (7.00%)
  1 (1.00%) low severe
  3 (3.00%) low mild
  1 (1.00%) high mild
  2 (2.00%) high severe
future_group_poll_latency/FutureGroup/Params { init_size: 1000, pct_ready_per_round: 0.2 }                                                                            
                        time:   [826.35 µs 828.36 µs 830.13 µs]
                        change: [+1.3397% +1.7624% +2.2118%] (p = 0.00 < 0.05)
                        Performance has regressed.
future_group_poll_latency/FuturesUnordered/Params { init_size: 1000, pct_ready_per_round: 0.2 }                                                                            
                        time:   [30.896 µs 31.006 µs 31.109 µs]
                        change: [-1.8062% -1.4034% -1.0374%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild
future_group_poll_latency/FutureGroup/Params { init_size: 1000, pct_ready_per_round: 1.0 }                                                                            
                        time:   [44.534 µs 44.610 µs 44.675 µs]
                        change: [+2.2008% +2.3854% +2.5792%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 25 outliers among 100 measurements (25.00%)
  4 (4.00%) low severe
  21 (21.00%) high severe
future_group_poll_latency/FuturesUnordered/Params { init_size: 1000, pct_ready_per_round: 1.0 }                                                                            
                        time:   [31.036 µs 31.122 µs 31.203 µs]
                        change: [-0.0946% +0.2297% +0.6345%] (p = 0.26 > 0.05)
                        No change in performance detected.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high severe

     Running benches/compare.rs (target/release/deps/compare-430c491dda6e2002)
Gnuplot not found, using plotters backend```

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -68,22 +68,30 @@ mod stream_group {
     }
 }
 mod future_group {
+    use std::fmt::{Debug, Display};
+    use std::time::{Duration, Instant};
+
     use criterion::async_executor::FuturesExecutor;
     use criterion::{black_box, criterion_group, BatchSize, BenchmarkId, Criterion};
+    use futures::channel::oneshot;
+    use futures::never::Never;
     use futures::stream::FuturesUnordered;
     use futures_concurrency::future::FutureGroup;
+    use futures_lite::future::yield_now;
     use futures_lite::prelude::*;
+    use itertools::Itertools;
+    use rand::{seq::SliceRandom, SeedableRng};
 
     use crate::utils::{make_future_group, make_futures_unordered};
     criterion_group! {
         name = future_group_benches;
         // This can be any expression that returns a `Criterion` object.
         config = Criterion::default();
-        targets = future_group_bench
+        targets = future_group_throughput_bench, future_group_latency_bench
     }
 
-    fn future_group_bench(c: &mut Criterion) {
-        let mut group = c.benchmark_group("future_group");
+    fn future_group_throughput_bench(c: &mut Criterion) {
+        let mut group = c.benchmark_group("future_group_poll_throughput");
         for i in [10, 100, 1000].iter() {
             group.bench_with_input(BenchmarkId::new("FutureGroup", i), i, |b, i| {
                 let setup = || make_future_group(*i);
@@ -113,6 +121,115 @@ mod future_group {
                 b.to_async(FuturesExecutor)
                     .iter_batched(setup, routine, BatchSize::SmallInput)
             });
+        }
+        group.finish();
+    }
+
+    /// This benchmark measures the latency between when futures become ready
+    /// and when their outputs appear on the [`FutureGroup`] stream.
+    ///
+    /// To test this, we:
+    /// - insert N pending futures to the [`FutureGroup`].
+    /// - until the [`FutureGroup`] is empty, we set some fraction of the
+    ///   pending futures to ready, then record how long it takes for their
+    ///   outputs to be produced from [`FutureGroup`]'s `Stream` impl.
+    /// - we sum the recorded durations for each of these rounds.
+    fn future_group_latency_bench(c: &mut Criterion) {
+        #[derive(Debug, Clone, Copy)]
+        struct Params {
+            init_size: usize,
+            pct_ready_per_round: f64,
+        }
+
+        impl Display for Params {
+            fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+                Debug::fmt(&self, f)
+            }
+        }
+
+        async fn routine<G>(
+            iters: u64,
+            Params {
+                init_size,
+                pct_ready_per_round,
+            }: Params,
+        ) -> Duration
+        where
+            G: Default + Stream + Unpin + Extend<oneshot::Receiver<Never>>,
+        {
+            let ready_per_round = ((init_size as f64) * pct_ready_per_round).max(1.) as usize;
+
+            let mut total_runtime = Duration::ZERO;
+
+            for _ in 0..iters {
+                // construct a set of oneshot::Receiver futures. These become
+                // ready once their Sender ends are dropped.
+                let (mut senders, mut group) = (0..init_size)
+                    .map(|_| oneshot::channel())
+                    .unzip::<_, _, Vec<_>, G>();
+
+                // shuffle our senders so from the FutureGroup's perspective,
+                // futures become ready in arbitrary order.
+                let mut rng = rand::rngs::StdRng::seed_from_u64(42);
+                senders.shuffle(&mut rng);
+
+                // poll once to set up all the wakers
+                assert!(futures::poll!(group.next()).is_pending());
+
+                while !senders.is_empty() {
+                    let num_completing = ready_per_round.min(senders.len());
+
+                    // drop some Senders. The corresponding Receiver futures
+                    // will become ready in the FutureGroup
+                    assert_eq!(senders.drain(..num_completing).count(), num_completing);
+
+                    // this isn't necessary now, but if we were using the tokio
+                    // runtime/oneshots, coming up for air regularly prevents
+                    // the budget system from inserting false pendings. this
+                    // also more closely emulates what would happen in a real
+                    // system (sender task yields to let executor poll receiver
+                    // task). though that shouldn't make a difference.
+                    yield_now().await;
+
+                    // measure the time it takes for all newly ready futures to
+                    // be produced from the FutureGroup stream.
+                    let recv_start = Instant::now();
+                    assert_eq!(
+                        (&mut group).take(num_completing).count().await,
+                        num_completing
+                    );
+                    total_runtime += recv_start.elapsed();
+                }
+            }
+
+            total_runtime
+        }
+
+        let mut group = c.benchmark_group("future_group_poll_latency");
+        for params in [10, 100, 1000]
+            .into_iter()
+            .cartesian_product([0.0001, 0.2, 1.0])
+            .map(|(init_size, pct_ready_per_round)| Params {
+                init_size,
+                pct_ready_per_round,
+            })
+        {
+            group.bench_with_input(
+                BenchmarkId::new("FutureGroup", params),
+                &params,
+                |b, &params| {
+                    b.to_async(FuturesExecutor)
+                        .iter_custom(|iters| routine::<FutureGroup<_>>(iters, params))
+                },
+            );
+            group.bench_with_input(
+                BenchmarkId::new("FuturesUnordered", params),
+                &params,
+                |b, &params| {
+                    b.to_async(FuturesExecutor)
+                        .iter_custom(|iters| routine::<FuturesUnordered<_>>(iters, params))
+                },
+            );
         }
         group.finish();
     }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,6 +49,7 @@ criterion = { version = "0.3", features = [
 ] }
 futures = "0.3.25"
 futures-time = "3.0.0"
+itertools = "0.12.1"
 lending-stream = "1.0.0"
 rand = "0.8.5"
 tokio = { version = "1.32.0", features = ["macros", "time", "rt-multi-thread"] }
--- a/src/future/future_group.rs
+++ b/src/future/future_group.rs
@@ -415,14 +415,22 @@ impl<F: Future> Stream for FutureGroup<F> {
     }
 }
 
-impl<F: Future> FromIterator<F> for FutureGroup<F> {
-    fn from_iter<T: IntoIterator<Item = F>>(iter: T) -> Self {
+impl<F: Future> Extend<F> for FutureGroup<F> {
+    fn extend<T: IntoIterator<Item = F>>(&mut self, iter: T) {
         let iter = iter.into_iter();
         let len = iter.size_hint().1.unwrap_or_default();
-        let mut this = Self::with_capacity(len);
+        self.reserve(len);
+
         for future in iter {
-            this.insert(future);
+            self.insert(future);
         }
+    }
+}
+
+impl<F: Future> FromIterator<F> for FutureGroup<F> {
+    fn from_iter<T: IntoIterator<Item = F>>(iter: T) -> Self {
+        let mut this = Self::new();
+        this.extend(iter);
         this
     }
 }
--- a/src/future/future_group.rs
+++ b/src/future/future_group.rs
@@ -59,7 +59,6 @@ use crate::utils::{PollState, PollVec, WakerVec};
 /// ```
 
 #[must_use = "`FutureGroup` does nothing if not iterated over"]
-#[derive(Default)]
 #[pin_project::pin_project]
 pub struct FutureGroup<F> {
     #[pin]
@@ -77,6 +76,12 @@ impl<T: Debug> Debug for FutureGroup<T> {
             .field("len", &self.len())
             .field("capacity", &self.capacity)
             .finish()
+    }
+}
+
+impl<T> Default for FutureGroup<T> {
+    fn default() -> Self {
+        Self::new()
     }
 }
 

### CODE REVIEW

File: benches/bench.rs [hunk 1]
```diff
@@ -116,6 +121,64 @@ mod future_group {
...
+    fn future_group_latency_bench(c: &mut Criterion) {
+        async fn routine<G>(iters: u64, size: usize) -> Duration
+        where
+            G: Default + Stream + Unpin + Extend<oneshot::Receiver<Never>>,
+        {
+            // fraction of `size` which represents the maximum number of
+            // futures which may all become ready between polls.
+            const MAX_READY_RATIO_PER_POLL: f64 = 0.1;
+            let max_ready_num_per_poll =
+                ((size as f64) * MAX_READY_RATIO_PER_POLL).max(1.) as usize;
```
soooch, line 134: sorry these names and explanations are awkward as hell. super open to suggestions.

benches/bench.rs
```diff
@@ -68,22 +68,27 @@ mod stream_group {
...
         name = future_group_benches;
         // This can be any expression that returns a `Criterion` object.
         config = Criterion::default();
-        targets = future_group_bench
+        targets = future_group_throughput_bench, future_group_latency_bench
     }
 
-    fn future_group_bench(c: &mut Criterion) {
-        let mut group = c.benchmark_group("future_group");
+    fn future_group_throughput_bench(c: &mut Criterion) {
```
soooch, line 90: throughput is maybe not the best name for this bench?

benches/bench.rs
```diff
@@ -116,6 +121,64 @@ mod future_group {
...
+                    let recv_start = Instant::now();
+                    assert_eq!(
+                        (&mut group).take(num_completing).count().await,
+                        num_completing
+                    );
+                    total_runtime += recv_start.elapsed();
+                }
+            }
+
+            total_runtime
```
yoshuawuyts, line 166: It took me a second to read how this works; it would be really helpful if you can annotate how the latency is being counted here.

From what I can tell, what we're doing is taking N rounds to complete all futures in the group. We count how long it took to complete each round, and add that to a total duration. The way we mark futures as complete is by inserting one-shot channels into the futures group.

Did I get that right?

benches/bench.rs
yoshuawuyts: Yeah, I hear you - I'm not sure what else to call it, so I think this is probably fine?

benches/bench.rs
```diff
@@ -116,6 +121,64 @@ mod future_group {
...
+
+            for _ in 0..iters {
+                let (mut senders, mut group) = (0..size)
+                    .map(|_| oneshot::channel())
+                    .unzip::<_, _, Vec<_>, G>();
+
+                let mut rng = rand::rngs::StdRng::seed_from_u64(42);
+
+                // senders will complete in arbitrary order
+                senders.shuffle(&mut rng);
```
yoshuawuyts, line 146: I don't know if this ends up making a huge difference, but I am a little uncomfortable with introducing internal RNG into the benchmark. In the benchmark results for the 1000 items we're seeing 1 in 4 samples being an outlier. That means either the benchmark itself or the system it was run on is producing unreliable results.

Can we make this reordering deterministic? Can we perhaps use the current sample counter to permute the senders list?

yoshuawuyts (changes requested):
Thanks so much for submitting this! - Despite not doing the best in these benchmarks yet, I think they're valuable because they allow us to learn where we have an opportunity to improve.

In principle I'm keen to land these; I mainly just have some concerns about the relatively high amount of outliers at the higher counts. If we can resolve that, I think we should be good!

Thank you!

yoshuawuyts:
@soooch By the way, I'd also be interested if you have any theories as to what might be causing this difference in latency? I know for example that in #137 we started dropping futures before we return their values, as a matter of correctness. I don't know whether `FuturesUnordered` does the same thing; but if they don't I could see that having a measurable impact on latency between futures completing and values being returned.

@soooch Also another question: it seems right now the test is hard-coded to taking 10 rounds. Have you noticed any differences if you increase or decrease the number of rounds?

benches/bench.rs
soooch: will get some more comments in there.

yoshuawuyts:
> my primary theory is that for each poll, all futures are iterated over regardless of whether they have been woken or not. iirc FuturesUnordered does some more complicated stuff there so it only iterates on woken futures. i think i saw an old PR on this repo that was trying to do something similar.

Hmmm, thank you for your answer - though that shouldn't be the case. We're being quite careful to only iterate over futures which are currently ready to be woken. We do iterate over all active keys, but that just walks a `BTreeSet`, which should be fairly cheap?:

https://github.com/yoshuawuyts/futures-concurrency/blob/dec5ce0c7f650aa6fccdcfe1e71147171402a5c4/src/future/future_group.rs#L363-L364

I'm quite surprised about the differences measured here; especially since on the existing benchmark we also measure time elapsed and the differences are effectively negligable. I'm hoping that by better understanding the way this benchmark is structured we can drill down further into why it performs differently for both APIs.

soooch:
> > my primary theory is that for each poll, all futures are iterated over regardless of whether they have been woken or not. iirc FuturesUnordered does some more complicated stuff there so it only iterates on woken futures. i think i saw an old PR on this repo that was trying to do something similar.
> 
> Hmmm, thank you for your answer - though that shouldn't be the case. We're being quite careful to only iterate over futures which are currently ready to be woken. We do iterate over all active keys, but that just walks a `BTreeSet`, which should be fairly cheap?:
> 
> https://github.com/yoshuawuyts/futures-concurrency/blob/dec5ce0c7f650aa6fccdcfe1e71147171402a5c4/src/future/future_group.rs#L363-L364
> 
> I'm quite surprised about the differences measured here; especially since on the existing benchmark we also measure time elapsed and the differences are effectively negligable. I'm hoping that by better understanding the way this benchmark is structured we can drill down further into why it performs differently for both APIs.

The`is_pending` and `clear_ready` are probably more expensive than the btree iteration. not sure if they’re expensive in an absolute sense though .But if it’s done for 1000 items when only a single item has woken up that should start to add up, right?

benches/bench.rs
soooch: Did verify that this usage of rand is deterministic

benches/bench.rs
soooch: @yoshuawuyts let me know if the new docs make this more clear

benches/bench.rs
soooch: also I removed the randomness from the "ready_per_round" count. That is now a parameter of the bench.

soooch:
> > > my primary theory is that for each poll, all futures are iterated over regardless of whether they have been woken or not. iirc FuturesUnordered does some more complicated stuff there so it only iterates on woken futures. i think i saw an old PR on this repo that was trying to do something similar.
> > 
> > 
> > Hmmm, thank you for your answer - though that shouldn't be the case. We're being quite careful to only iterate over futures which are currently ready to be woken. We do iterate over all active keys, but that just walks a `BTreeSet`, which should be fairly cheap?:
> > https://github.com/yoshuawuyts/futures-concurrency/blob/dec5ce0c7f650aa6fccdcfe1e71147171402a5c4/src/future/future_group.rs#L363-L364
> > 
> > I'm quite surprised about the differences measured here; especially since on the existing benchmark we also measure time elapsed and the differences are effectively negligable. I'm hoping that by better understanding the way this benchmark is structured we can drill down further into why it performs differently for both APIs.
> 
> The`is_pending` and `clear_ready` are probably about as expensive as the btree iteration. they’re not expensive in an absolute sense. But if they're done for 1000 items when only a single item has woken up that should start to add up, right?
> 
> Take a look at the `ready_to_run_queue` in `FuturesUnordered`

I think I've confirmed this theory to myself. the following change brings FutureGroup to parity with FuturesUnordered for this benchmark.

We are still more algorithmically complex than FuturesUnordered, but iter_ones on BitVec is sufficiently well optimized that we're still faster than FuturesUnordered until the test size gets to around 10000 futures.

```diff
diff --git a/src/future/future_group.rs b/src/future/future_group.rs
index bcb4fa4..45b48cf 100644
--- a/src/future/future_group.rs
+++ b/src/future/future_group.rs
@@ -1,4 +1,3 @@
-use alloc::collections::BTreeSet;
 use core::fmt::{self, Debug};
 use core::ops::{Deref, DerefMut};
 use core::pin::Pin;
@@ -65,7 +64,6 @@ pub struct FutureGroup<F> {
     futures: Slab<F>,
     wakers: WakerVec,
     states: PollVec,
-    keys: BTreeSet<usize>,
     capacity: usize,
 }
 
@@ -115,7 +113,6 @@ impl<F> FutureGroup<F> {
             futures: Slab::with_capacity(capacity),
             wakers: WakerVec::new(capacity),
             states: PollVec::new(capacity),
-            keys: BTreeSet::new(),
             capacity,
         }
     }
@@ -190,10 +187,9 @@ impl<F> FutureGroup<F> {
     /// # })
     /// ```
     pub fn remove(&mut self, key: Key) -> bool {
-        let is_present = self.keys.remove(&key.0);
+        let is_present = self.futures.try_remove(key.0).is_some();
         if is_present {
             self.states[key.0].set_none();
-            self.futures.remove(key.0);
         }
         is_present
     }
@@ -215,7 +211,7 @@ impl<F> FutureGroup<F> {
     /// # })
     /// ```
     pub fn contains_key(&mut self, key: Key) -> bool {
-        self.keys.contains(&key.0)
+        self.futures.contains(key.0)
     }
 
     /// Reserves capacity for `additional` more futures to be inserted.
@@ -270,7 +266,6 @@ impl<F: Future> FutureGroup<F> {
         }
 
         let index = self.futures.insert(future);
-        self.keys.insert(index);
 
         // Set the corresponding state
         self.states[index].set_pending();
@@ -293,7 +288,6 @@ impl<F: Future> FutureGroup<F> {
         // SAFETY: inserting a value into the futures slab does not ever move
         // any of the existing values.
         let index = unsafe { this.futures.as_mut().get_unchecked_mut() }.insert(future);
-        this.keys.insert(index);
         let key = Key(index);
 
         // If our slab allocated more space we need to
@@ -365,8 +359,8 @@ impl<F: Future> FutureGroup<F> {
         // single futures. Either to read from them or to drop them.
         let futures = unsafe { this.futures.as_mut().get_unchecked_mut() };
 
-        for index in this.keys.iter().cloned() {
-            if states[index].is_pending() && readiness.clear_ready(index) {
+        while let Some(index) = readiness.take_next_ready() {
+            if states[index].is_pending() {
                 // unlock readiness so we don't deadlock when polling
                 #[allow(clippy::drop_non_drop)]
                 drop(readiness);
@@ -398,12 +392,6 @@ impl<F: Future> FutureGroup<F> {
             }
         }
 
-        // Now that we're no longer borrowing `this.keys` we can remove
-        // the current key from the set
-        if let Poll::Ready(Some((key, _))) = ret {
-            this.keys.remove(&key.0);
-        }
-
         ret
     }
 }
diff --git a/src/utils/wakers/vec/readiness_vec.rs b/src/utils/wakers/vec/readiness_vec.rs
index d7a8640..d392c6b 100644
--- a/src/utils/wakers/vec/readiness_vec.rs
+++ b/src/utils/wakers/vec/readiness_vec.rs
@@ -65,6 +65,12 @@ impl ReadinessVec {
         self.ready_count > 0
     }
 
+    pub(crate) fn take_next_ready(&mut self) -> Option<usize> {
+        let index = self.readiness_list.iter_ones().next()?;
+        self.clear_ready(index);
+        Some(index)
+    }
+
     /// Access the parent waker.
     #[inline]
     pub(crate) fn parent_waker(&self) -> Option<&Waker> {
```

benches/bench.rs
```diff
@@ -116,6 +124,116 @@ mod future_group {
...
+
+    /// This benchmark aims to measure how long it takes for the results of
+    /// futures that have become ready to be produced by the [`FutureGroup`]
+    /// stream.
+    ///
+    /// To test this, we:
+    /// - insert N pending futures to the [`FutureGroup`].
+    /// - until the [`FutureGroup`] is empty, we set some fraction of the
+    ///   pending futures to ready, then record how long it takes for all newly
+    ///   ready futures to be produced from [`FutureGroup`]'s `Stream` impl.
```
soooch, line 136:
```suggestion
    /// - until the [`FutureGroup`] is empty, we set some fraction of the
    ///   pending futures to ready, then record how long it takes for their
    ///   outputs to be produced from [`FutureGroup`]'s `Stream` impl.
```

benches/bench.rs
```diff
@@ -116,6 +124,116 @@ mod future_group {
         }
         group.finish();
     }
+
+    /// This benchmark aims to measure how long it takes for the results of
+    /// futures that have become ready to be produced by the [`FutureGroup`]
+    /// stream.
```
soooch, line 130:
```suggestion
    /// This benchmark measures the latency between futures becoming ready
    /// and their outputs appearing on the [`FutureGroup`] stream.
```

soooch:
> Thanks so much for submitting this! - Despite not doing the best in these benchmarks yet, I think they're valuable because they allow us to learn where we have an opportunity to improve.
> 
> In principle I'm keen to land these; I mainly just have some concerns about the relatively high amount of outliers at the higher counts. If we can resolve that, I think we should be good!
> 
> Thank you!

So it appears I have a significant number of outliers for nearly every bench included in this crate on my machine. This new benchmark is about on par with the rest.

soooch:
@yoshuawuyts this may have fallen through the cracks. I believe I've addressed your initial concerns.

yoshuawuyts (approved):
Thanks again!

PR #181, opened 2024-04-18
Add checks for msrv and semver

### LINKED ISSUE #180: Declare minimum support rust version in Cargo manifest
Today, this crate doesn't declare the [msrv](https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field) in the `Cargo.toml`.

This can lead to more confusing errors when trying to compile with an older Rust compiler, and when trying to find the right release that would work with an older Rust compiler.

This would likely need a pipeline change as well, so the CI verifies the declared msrv. There seems to be notes on that [here](https://doc.rust-lang.org/cargo/guide/continuous-integration.html#verifying-rust-version).

matheus-consoli:
declare the Minimum Supported Rust Version and add a check for it in CI. the msrv was identified using `cargo-msrv`.

i also took the opportunity to fill the `keywords` and `categories` on the cargo manifest, and, more importantly, add a CI step for `cargo-semver-checks`. the crate is gaining more features and users (!!), it's important not to break semver expectations.

Closes #180

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -91,3 +91,11 @@ jobs:
 
     - name: Docs
       run: cargo doc
+
+  semver-checks:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Check semver
+        uses: obi1kenobi/cargo-semver-checks-action@v2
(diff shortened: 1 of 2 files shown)


yoshuawuyts (approved):
This is great; thank you!

PR #187, opened 2024-05-16
fix pin violation in much of concurrentstream

### LINKED ISSUE #182: ConcurrentStream usage with tokio leads to ACCESS_VIOLATION
Tried to use concurrent streams to sleep in parallel with tokio:
```rs
use core::time::Duration;
use futures_concurrency::prelude::*;
use tokio;
use futures::prelude::*;

#[tokio::main(flavor = "current_thread")]
async fn main() {
    let s1 = stream::iter([1, 2]);
    let s2 = stream::iter([1, 2]);
    let s3 = stream::iter([1, 2]);

    (s1, s2, s3)
        .chain()
        .co()
        .map(|x| async move {
            tokio::time::sleep(Duration::from_secs(x as _)).await;
            ()
        })
        .for_each(|_| async {})
        .await;
}
```

But sometimes I get crash:
```sh
> cargo run
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.15s
     Running `target\debug\rstestss.exe`
error: process didn't exit successfully: `target\debug\rstestss.exe` (exit code: 0xc0000005, STATUS_ACCESS_VIOLATION)
```

Without "current_thread" flavor program just freeze

Other runtimes work fine:
async_std
```rs
use core::time::Duration;
use async_std;
use futures_concurrency::prelude::*;
use futures::prelude::*;

#[async_std::main]
async fn main() {
    let s1 = stream::iter([1, 2]);
    let s2 = stream::iter([1, 2]);
    let s3 = stream::iter([1, 2]);

    (s1, s2, s3)
        .chain()
        .co()
        .map(|x| async move {
            async_std::task::sleep(Duration::from_secs(x)).await;
            ()
        })
        .for_each(|_| async {})
        .await;
}
```

smol
```rs
use core::time::Duration;
use futures_concurrency::prelude::*;
use futures::prelude::*;

async fn main_async() {
    let s1 = stream::iter([1, 2]);
    let s2 = stream::iter([1, 2]);
    let s3 = stream::iter([1, 2]);

    (s1, s2, s3)
        .chain()
        .co()
        .map(|x| async move {
            smol::Timer::after(Duration::from_secs(x)).await;
            ()
        })
        .for_each(|_| async {})
        .await;
}

fn main() {
    smol::block_on(main_async());
}
```

Also tokio runtime with smol Timer works fine:
```rs
use core::time::Duration;
use futures_concurrency::prelude::*;
use tokio;
use futures::prelude::*;

#[tokio::main(flavor = "current_thread")]
async fn main() {
    let s1 = stream::iter([1, 2]);
    let s2 = stream::iter([1, 2]);
    let s3 = stream::iter([1, 2]);

    (s1, s2, s3)
        .chain()
        .co()
        .map(|x| async move {
            // tokio::time::sleep(Duration::from_secs(x as _)).await;
            smol::Timer::after(Duration::from_secs(x)).await;
            ()
        })
        .for_each(|_| async {})
        .await;
}
```

Is that Tokio issue?

conradludgate:
Fixes #182

I developed the crate `futures-buffered` a couple years ago with the intent of being a more efficient 'FuturesUnordered' type than the one in `futures_util`.

The core of the unsoundness in #182 are the pin-violations caused by FutureGroup slab reallocs. `futures-buffered` solves this by using a 'triangular array' pattern, where each subsequent realloc is actually a new slab.

Note: the unsoundness still exists in FutureGroup. `futures-buffered` does not return keys for entries (although that is not a bad idea) so the API is not compatible.

A quick soundness fix for FutureGroup would be to make `Stream` require `F: Unpin`, albeit this is a breaking change.

Even if you do not merge this change, I hope it serves as a nice inspiration for you :)

### DIFF
--- a/src/future/future_group.rs
+++ b/src/future/future_group.rs
@@ -274,6 +274,38 @@ impl<F: Future> FutureGroup<F> {
         Key(index)
     }
 
+    #[allow(unused)]
+    /// Insert a value into a pinned `FutureGroup`
+    ///
+    /// This method is private because it serves as an implementation detail for
+    /// `ConcurrentStream`. We should never expose this publicly, as the entire
+    /// point of this crate is that we abstract the futures poll machinery away
+    /// from end-users.
+    pub(crate) fn insert_pinned(self: Pin<&mut Self>, future: F) -> Key
+    where
+        F: Future,
+    {
+        let mut this = self.project();
+        // SAFETY: inserting a value into the futures slab does not ever move
+        // any of the existing values.
+        let index = unsafe { this.futures.as_mut().get_unchecked_mut() }.insert(future);
+        this.keys.insert(index);
+        let key = Key(index);
+
+        // If our slab allocated more space we need to
+        // update our tracking structures along with it.
+        let max_len = this.futures.as_ref().capacity().max(index);
+        this.wakers.resize(max_len);
+        this.states.resize(max_len);
+
+        // Set the corresponding state
+        this.states[index].set_pending();
+        let mut readiness = this.wakers.readiness();
+        readiness.set_ready(index);
+
+        key
+    }
+
     /// Create a stream which also yields the key of each item.
     ///
     /// # Example
--- a/src/future/future_group.rs
+++ b/src/future/future_group.rs
@@ -274,37 +274,6 @@ impl<F: Future> FutureGroup<F> {
         Key(index)
     }
 
-    /// Insert a value into a pinned `FutureGroup`
-    ///
-    /// This method is private because it serves as an implementation detail for
-    /// `ConcurrentStream`. We should never expose this publicly, as the entire
-    /// point of this crate is that we abstract the futures poll machinery away
-    /// from end-users.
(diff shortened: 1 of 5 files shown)


src/future/future_group.rs
```diff
@@ -274,37 +274,6 @@ impl<F: Future> FutureGroup<F> {
         Key(index)
     }
 
-    /// Insert a value into a pinned `FutureGroup`
```
yoshuawuyts, line 277: can we keep these methods, but mark them as `#[allow(unused)]` for the time being?

yoshuawuyts (approved):
Overall this looks great, thank you so much! - I'd like to eventually update our own `{Future,Stream}Group` types to not have the issues. And actually would prefer to keep external deps to a minimum.

But I'm also recognizing that we shouldn't let perfect be the enemy of good, and this actually fixes our issues straight away - so I'm keen to accept this patch and then figure out where we go after that.

If you could revert deleting the methods in `future_group`, I think we should be good to merge! Thank you so much!

src/future/future_group.rs
conradludgate: Of course

PR #192, opened 2024-08-01
fix: `clippy::doc_lazy_continuation`

crepererum:
See https://rust-lang.github.io/rust-clippy/master/index.html#/doc_lazy_continuation

Was introduced recently, and now makes PRs fail.


yoshuawuyts (approved):
this is great, thank you!

PR #194, opened 2024-09-20
Reference flush instead of finnish in ConsumerState::Break doc

tyilo:


yoshuawuyts (approved):
nice catch; thank you!

PR #196, opened 2024-10-18
fix no_std

Easyoakland:
Fixes `no_std` after after futures-buffered accepts https://github.com/conradludgate/futures-buffered/pull/9.

`futures-lite` has `std` and `alloc` features by default which had to be disabled unless the same feature is enabled in this crate.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,13 +28,13 @@ harness = false
 
 [features]
 default = ["std"]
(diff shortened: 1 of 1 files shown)


Easyoakland:
Clippy checks don't look relevant to this PR.

yoshuawuyts (approved):
thank you!

PR #204, opened 2025-01-27
Fix RaceOk tuple and array impls

jstarks:
Both the tuple and array implementations of `RaceOk` would poll futures that had already been polled to completion. Fix these by tracking per-future state.

Additionally, the array implementation failed to drop any accumulated errors in the success case. Add a drop implementation to fix this.

### DIFF
--- a/src/future/race_ok/array/mod.rs
+++ b/src/future/race_ok/array/mod.rs
@@ -1,6 +1,7 @@
 use super::RaceOk as RaceOkTrait;
 use crate::utils::array_assume_init;
 use crate::utils::iter_pin_mut;
+use crate::utils::PollArray;
 
 use core::array;
 use core::fmt;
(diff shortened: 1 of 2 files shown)


yoshuawuyts:
Thank you for filing this @jstarks! I want to acknowledge I've seen this, and I will review it within the next 24hrs. The problems you're highlighting here seem important to address, and I want to make sure this gets a proper review.

jstarks:
Thanks, Yosh!

yoshuawuyts:
Okay, reviewed! It took me a second to page this back in. I see what you're doing now, and it matches the pattern we're using in the other combinators too. Tldr for posterity: `MaybeUninit<T>` types don't have their destructors run automatically, so we need to track their state and manually run their destructors.

Thank you so much for filing and fixing this @jstarks!

PR #206, opened 2025-03-05
Don't include default features of slab

emil-e:
`slab` includes the `std` feature by default which won't work for `no_std` projects. I can't really reproduce the problem by just building the project by itself but when pulling in `futures-concurrency` in my `no_std`/`embassy` project, I get errors such as:
```
cannot find type `Option` in this scope
not found in this scope   
```

Fix is simply to exclude default features from `slab`. Seems to work even when `std` features is enabled for `futures-concurrency`.


emil-e:
The Clippy failure is (obviously) unrelated to this change and also seems like a false positive.

emil-e:
@yoshuawuyts or someone else, any input on this?

PR #207, opened 2025-04-02
fix: make `futures-buffered` optional dependency in respect to `alloc`

Ddystopia:
`futures-buffered` is only used in `concurrent_stream`, which is feature gated to `feature = "alloc"`. But dependency is not optional, thus compiler still requires you to devine a global allocator.


Ddystopia:
CI failures are unrelated:
>  error: package `half v2.5.0` cannot be built because it requires rustc 1.81 or newer, while the currently active rustc version is 1.75.0

PR #208, opened 2025-06-10
Add up to 15 join/try_join tuple items

TTWNO:
Hello! Thank you so much for your create!

I work on a screen reader called Odilia.
When we get a new item we have not seen before, we asynchronously grab all the fields of the item at once.
Since each field is a different type, we use `try_join!`.

We recently added a few new fields that go over the limit of the `(...).try_join` implementation.
This PR makes the maximum 15 futures instead of 12.


yoshuawuyts (approved):
Thank you!

Issue #188, opened 2024-06-09
Fix pin violation in `FutureGroup`


yoshuawuyts:
Following up on #187, this problem still persists in `FutureGroup` itself. I see two options for resolving this and moving forward with a permanent resolution:

1. We keep the existing architecture for `FutureGroup` based on a slab structure, but replace it with a more efficient implementation such as the [`beton` crate](https://github.com/yoshuawuyts/beton).
2. Change the internal architecture of `FutureGroup` to more closely match `FuturesUnordered`. I haven't fully paged in how that works yet, but we know it doesn't have the same issues we have, so that might be a lower-risk solution.

cc/ @conradludgate Given you've worked on [futures-buffered](https://docs.rs/futures-buffered/latest/futures_buffered/), and probably know the `FuturesUnordered` internals better than I do, I'd be interested in your thoughts here.

conradludgate:
The slab implementation in `beton` would still exhibit the same unsoundness that `FutureGroup` currently does. The core of the issue is that when you inevitably need to realloc the slab, it will move all items that were allocated - this will violate any `Pin` requirements. You can over come this by requiring `F: Unpin`, but that's quite limiting.

I'll try give a simplified overview of the internals of `futures` and `futures-buffered` and how they overcome this constraint.

## `futures::future::FuturesUnordered`

This implementation uses an intrusive linked list construction. Each future spawned in the group is allocated in it's own arc, and linked into the ready list. When the future is polled and not ready, it is unlinked and linked into the idle list. The task doubles as it's own waker through `ArcWake`. When woken, the task unlinks itself from the idle list and links into the ready list. It then wakes the waker that is stored in the `FuturesUnordered`.

This implementation is not unlike how most executors work, although it's not the most efficient. Tokio, for instance, has an additional fixed length ring buffer that it keeps a batch of ready tasks in - since popping from a ring buffer is cheaper than removing from a cache-unfriendly linked list.

## `futures_buffered::FuturesUnordered`

This implementation uses a "triangular array" to amortise some of the allocation cost and potentially allow for allocation re-use. (Similar properties as a Vec<T>, except it is non-contiguous and never re-allocates, which preserves pin properties). This is also what the `thread_local` crate uses. Simply put, it's a `Vec<Box<[T]>>` where each subsequent box-slice is double the length of the slice before it:

```
[[1], [2, 3], [4, 5, 6, 7], [8, 9, 10, 11, 12, 13, 14, 15]]
```

[...]

PR #222, opened 2026-01-13
Fix pinned drop violations; make types Unpin where practical

A248:
I was writing a library of my own and decided to test this one. Although I had seen that #188 described pinning problems, I only discovered the extent of this myself, when this library began failing my code's anti-move checks.

This PR shores up the pin-on-drop problem. Pinned types have to be dropped in the same place, and ChunkedVec wasn't doing this.

### Changes summary

* FutureGroup and StreamGroup are now `Unpin` types. At heart, they kinda always were (see below)
* ChunkedVec is now built around pinning, fulfilling its destiny as a data structure made specifically for pinnable data
* Brief changes to tests, to catch the bug that spawned this PR to begin with

### Unpinned collections

Properly speaking, neither FutureGroup nor StreamGroup need to be pinned. The futures/streams are placed in allocations managed by ChunkedVec, which is responsible for keeping the futures/streams pinned. This has been true. In fact, this means the entire pin_project/`#[pin]` management done by these types was superfluous. FutureGroup and StreamGroup don't store any pinnable data inline, not even when the implementation used Slab.

Going forward, these types can be safely exposed as Unpin. Even if their implementation changes again, FutureGroup and StreamGroup are never going to store their streams/futures on stack. Only inline storage makes pinning self-referential data meaningful; heap storage will always be stable. Thus, making them Unpin is totally safe from the perspective of future compatibility.

### Growing streams and polling them too

As a side effect of making FutureGroup and StreamGroup `Unpin` types, they can now have items added to them even after they've been pinned. This was already the case for `Unpin` items, but now it will be true for `!Unpin` types as well.

This was alluded to in #188: by requiring the group to be pinned in order to be polled, the size becomes frozen and new insertions are impossible. I didn't investigate the previous version of futures-concurrency very much, but "freezing" the collection might be the only reason it used pinning.

That said, abandoning that design is surely for the better, and I don't think future implementations of this crate would ever need to bring it back. It was more of a side effect and a quirk of how the crate was implemened.

### DIFF
--- a/src/concurrent_stream/for_each.rs
+++ b/src/concurrent_stream/for_each.rs
@@ -12,7 +12,6 @@ use core::sync::atomic::{AtomicUsize, Ordering};
 use core::task::{ready, Context, Poll};
 
 // OK: validated! - all bounds should check out
-#[pin_project]
 pub(crate) struct ForEachConsumer<FutT, T, F, FutB>
 where
     FutT: Future<Output = T>,
@@ -21,11 +20,18 @@ where
 {
     // NOTE: we can remove the `Arc` here if we're willing to make this struct self-referential
     count: Arc<AtomicUsize>,
-    #[pin]
     group: FutureGroup<ForEachFut<F, FutT, T, FutB>>,
     limit: usize,
     f: F,
     _phantom: PhantomData<(T, FutB)>,
+}
+
+impl<FutT, T, F, FutB> Unpin for ForEachConsumer<FutT, T, F, FutB>
+where
+    FutT: Future<Output = T>,
+    F: Fn(T) -> FutB,
+    FutB: Future<Output = ()>,
+{
 }
 
 impl<A, T, F, B> ForEachConsumer<A, T, F, B>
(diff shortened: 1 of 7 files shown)


A248:
I don't know why, however, MIRI is very slow to run the pin_safety tests. If this is not related to this PR, feel free to disregard.

```
Running tests/pin_safety.rs (target/miri/x86_64-unknown-linux-gnu/debug/deps/pin_safety-385fd5db276d1151)
running 4 tests
test future_group_no_move_on_repeated_growth ... ok
test stream_group_no_move_on_growth ... ok
test stream_group_no_move_on_repeated_growth ... ok
test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 67.99s
```

yoshuawuyts:
Thanks for submitting this patch! I thing @matheus-consoli is probably the right person to review this, which is why I've assigned them to it. However if they don't get around to it by the end of next week, please feel free to ping me and I'll do my best to review this.

matheus-consoli (approved):
Hey, thank you so much for fixing this! Everything looks great

thinety:
Hi,

I'm sorry if this is the wrong place to ask, but is there an estimate for when the next release that includes this fix will be published?

At `$WORK`, I hit a segfault when using the `StreamGroup::remove` API, and it turns out it's fixed by the changes introduced by this PR. Right now, we're using a patch pointing directly to the main branch, but it would be nice to use an official release instead.

Thanks for the work on this library!

PR #223, opened 2026-01-25
remove pin_project'ion on `Keyed`

matheus-consoli:
`Keyed` doesn't need to `#[pin]` the group anymore, because the underlying types are `Unpin.`


yoshuawuyts (approved):
This is great, ty!

